PR #23, opened 2020-06-15
Feature/pro 33

ctucker3:
Implements changes to check the cookies on successful login and looks for a cookie named `redirect-login`. If the cookie is present, and has a valid url as its value, the user is redirected to that url on successful login. If the cookie is not present or is not a valid url, the user is redirected to the default url set in the config. 

Should be reviewed with PRO-33 on bi-web.

### DIFF
--- a/src/main/java/org/breedinginsight/api/auth/AuthServiceLoginHandler.java
+++ b/src/main/java/org/breedinginsight/api/auth/AuthServiceLoginHandler.java
@@ -37,9 +37,7 @@ import org.breedinginsight.services.UserService;
 import javax.inject.Inject;
 import javax.inject.Singleton;
 import java.io.UnsupportedEncodingException;
-import java.net.URI;
-import java.net.URISyntaxException;
-import java.net.URLDecoder;
+import java.net.*;
 import java.nio.charset.StandardCharsets;
 import java.util.Iterator;
 import java.util.List;
@@ -96,15 +94,16 @@ public class AuthServiceLoginHandler extends JwtCookieLoginHandler {
                 if (request.getCookies().contains(loginSuccessUrlCookieName)){
                     Cookie loginSuccessCookie = request.getCookies().get(loginSuccessUrlCookieName);
                     String returnUrl = loginSuccessCookie.getValue();
-                    if (isValidURI(returnUrl)){
-                        try {
-                            locationUrl = URLDecoder.decode(returnUrl, StandardCharsets.UTF_8.name());
-                        } catch (UnsupportedEncodingException e){}
-                    }
+                    try {
+                        returnUrl = URLDecoder.decode(returnUrl, StandardCharsets.UTF_8.name());
+                        if (isValidURL(returnUrl)){
+                            locationUrl = returnUrl;
+                        }
+                    } catch (UnsupportedEncodingException e){}
                 }
             }
 
-            URI location = new URI(locationUrl).normalize();
+            URI location = new URI(locationUrl);
 
             MutableHttpResponse mutableHttpResponse = HttpResponse.seeOther(location);
 
@@ -126,11 +125,11 @@ public class AuthServiceLoginHandler extends JwtCookieLoginHandler {
         return super.loginFailed(authenticationFailed);
     }
 
-    private Boolean isValidURI(String url) {
+    private Boolean isValidURL(String urlString) {
         try {
-            new URI(url);
-            return true;
-        } catch (URISyntaxException e){
+            URL url = new URL(urlString);
+            return "https".equals(url.getProtocol()) || "http".equals(url.getProtocol());
+        } catch (MalformedURLException e){
             return false;
         }
     }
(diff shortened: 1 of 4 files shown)


nickpalladino:
[PRO-33 - Reauth User](https://trello.com/c/csSlPBi4/226-pro-33-reauth-user)

ctucker3:
> I got both bi-api and bi-web running. Works like a charm! One question: I can't seem to get the `redirect-login` cookie to set regardless of the page I've visited. When should I be expecting to see that cookie?
> I get only `phylo-token` on any page (Chrome v. 83.0.4, MacOS 10.14.6).

The `redirect-login` cookie will be set when you are not logged in and try to visit a page. You should be redirected to the home page immediately. There you will see the `redirect-login` cookie. The `redirect-login` cookie is removed each time you navigate to a page so that it doesn't hang about. So, you will only see it on the home page after a redirect. 

Is the redirect flow for the navigating to a protected page from an external link working for you?

src/main/java/org/breedinginsight/api/auth/AuthServiceLoginHandler.java
```diff
@@ -43,6 +50,7 @@
 
     @Inject
     private UserService userService;
+    private String loginSuccessUrlCookieName = "redirect-login";
```
nickpalladino, line 53: Do we want the cookie name in the config file? You could then inject it here.

src/main/java/org/breedinginsight/api/auth/AuthServiceLoginHandler.java
```diff
@@ -75,10 +83,54 @@ public HttpResponse loginSuccess(UserDetails userDetails, HttpRequest<?> request
...
+                    } catch (UnsupportedEncodingException e){}
+                }
+            }
+
+            URI location = new URI(locationUrl);
+
+            MutableHttpResponse mutableHttpResponse = HttpResponse.seeOther(location);
+
+            Cookie cookie;
+            for(Iterator var4 = cookies.iterator(); var4.hasNext(); mutableHttpResponse = mutableHttpResponse.cookie(cookie)) {
```
nickpalladino, line 111: Curious about these var4 and var6 variable names, maybe could be something a little more informative.

src/main/java/org/breedinginsight/api/auth/AuthServiceLoginHandler.java
```diff
@@ -75,10 +83,54 @@ public HttpResponse loginSuccess(UserDetails userDetails, HttpRequest<?> request
...
+                HttpRequest<Object> request = requestOptional.get();
+                if (request.getCookies().contains(loginSuccessUrlCookieName)){
+                    Cookie loginSuccessCookie = request.getCookies().get(loginSuccessUrlCookieName);
+                    String returnUrl = loginSuccessCookie.getValue();
+                    try {
+                        returnUrl = URLDecoder.decode(returnUrl, StandardCharsets.UTF_8.name());
+                        if (isValidURL(returnUrl)){
+                            locationUrl = returnUrl;
+                        }
+                    } catch (UnsupportedEncodingException e){}
```
nickpalladino, line 102: May be useful to have a log entry instead of silent exception.

src/main/java/org/breedinginsight/api/auth/AuthServiceLoginHandler.java
```diff
@@ -75,10 +83,54 @@ public HttpResponse loginSuccess(UserDetails userDetails, HttpRequest<?> request
...
+            if (requestOptional.isPresent()){
+                HttpRequest<Object> request = requestOptional.get();
+                if (request.getCookies().contains(loginSuccessUrlCookieName)){
+                    Cookie loginSuccessCookie = request.getCookies().get(loginSuccessUrlCookieName);
+                    String returnUrl = loginSuccessCookie.getValue();
+                    try {
+                        returnUrl = URLDecoder.decode(returnUrl, StandardCharsets.UTF_8.name());
+                        if (isValidURL(returnUrl)){
+                            locationUrl = returnUrl;
+                        }
```
nickpalladino, line 101: May be useful to have a log entry for the invalid url case.

src/main/java/org/breedinginsight/api/auth/AuthServiceLoginHandler.java
ctucker3: This is just copied and pasted from the micronaut loginSuccessCookies method. Not a problem to change them for our use though.

ctucker3: changed and pushed

src/main/java/org/breedinginsight/api/auth/AuthServiceLoginHandler.java
ctucker3: changed and pushed

src/main/java/org/breedinginsight/api/auth/AuthServiceLoginHandler.java
ctucker3: changed and pushed

src/main/java/org/breedinginsight/api/auth/AuthServiceLoginHandler.java
ctucker3: changed and pushed

PR #24, opened 2020-06-22
Feature/pro 53

ctucker3:
Adds the program roles to the user controller return data. The program roles object that is returned now includes the program object that the role is referring to. Inactive program users will be returned for a user, program users that belong to an inactive program will not be returned. 

This will be useful in the future feature of permission checks in bi-api on the program level. 

Also includes updates to the UserControllerIntegrationTest to update the test data creation and pulling using fannypack. 

## Dependencies

This feature is paired with PRO-53 of bi-web. 

## Review

- Tests are adjusted properly for new user controller return objects. 
- Works with bi-web PRO-53. 
- Normal code review

### DIFF
--- a/docs/openapi.yaml
+++ b/docs/openapi.yaml
@@ -949,7 +949,7 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/programUserResponse'
-              example:
+              examples:
                 programUserExample2:
                   $ref: '#/components/examples/singleProgramUser'
         "400":
@@ -2486,6 +2486,18 @@ components:
           description: Array of roles
           items:
             $ref: '#/components/schemas/role'
+        program:
+          type: object
+          properties:
+            id:
+              type: string
+              format: uuid
+            name:
+              type: string
+            abbreviation:
+              type: string
+            objective:
+              type: string
     program:
       type: object
       properties:
@@ -3532,6 +3544,11 @@ components:
               roles:
                 - id: e2d6c5c3-4064-4afe-abd6-0af90fbf2245
                   domain: Breeder
+              program:
+                id: 37935efa-4c0f-4269-8669-f78d21b26f04
+                name: Test Program
+                abbreviation: test,
+                objective: To test things
             - createdAt: "2020-02-09T09:57:35-04:00"
               updatedAt: "2020-02-09T09:57:35-04:00"
               createdByUser:
@@ -3548,7 +3565,12 @@ components:
                 email: tim.smith@mail.com
               roles:
                 - id: e2d6c5c3-4064-4afe-abd6-0af90fbf2245
-                  domain: Breeder
+                  domain: Breeder'
+              program:
+                id: 37935efa-4c0f-4269-8669-f78d21b26f04
+                name: Test Program
+                abbreviation: test,
+                objective: To test things
     singleProgramUser:
       value:
         metadata:
@@ -3577,6 +3599,11 @@ components:
           roles:
             - id: e2d6c5c3-4064-4afe-abd6-0af90fbf2245
               domain: Field Manager
+          program:
+            id: 37935efa-4c0f-4269-8669-f78d21b26f04
+            name: Test Program
+            abbreviation: test,
+            objective: To test things
     arrayOfCountries:
       value:
         metadata:
--- a/docs/openapi.yaml
+++ b/docs/openapi.yaml
@@ -74,10 +74,36 @@ paths:
                         domain: "admin"
                       }
                       ]
-                      activePrograms: [
+                      programRoles: [
                       {
-                        id: eac1f9ba-c0da-48aa-bac6-16e178dbbbcc,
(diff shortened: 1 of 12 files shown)


src/main/java/org/breedinginsight/model/User.java
```diff
@@ -63,6 +69,7 @@ public static User parseSQLRecord(Record record, @NotNull BiUserTable tableName)
                 .name(record.getValue(tableName.NAME))
                 .email(record.getValue(tableName.EMAIL))
                 .systemRoles(new ArrayList<>())
+                .programRoles(new ArrayList<>())
                 .active(record.getValue(tableName.ACTIVE))
```
nickpalladino, line 73: There is a mismatch with the active field between here and the api docs. The response here includes it but the api docs schema and example doesn't.

src/main/java/org/breedinginsight/daos/UserDAO.java
```diff
@@ -83,36 +84,77 @@ public UserDAO(Configuration config, DSLContext dsl) {
...
+                Optional<SystemRole> systemRoleExists = Utilities.findInList(existingUser.getSystemRoles(),
+                        systemRole, SystemRole::getId);
+                if (!systemRoleExists.isPresent()){
+                    existingUser.addRole(systemRole);
+                }
             }
+
+            // Add our program
+            if (programUser.getProgramId() != null){
+                Optional<ProgramUser> programUserExists = Utilities.findInList(existingUser.getProgramRoles(),
```
nickpalladino, line 127: When the program is deactivated a program role is returned but without the roles array which I assume is what you want? I didn't see a test case for this scenario.

ctucker3: I'll have to fix this. The intended functionality is for the program role to not be returned if the program is inactive.

src/main/java/org/breedinginsight/model/User.java
ctucker3: changed and pushed

src/main/java/org/breedinginsight/daos/UserDAO.java
ctucker3: I pushed up a fix for this so if the program is inactive, the program role is not returned. I had to add onto our Jooq parsing pattern because our previous flow did not support inner select queries, which I needed for this case. 

I ended up needing to generate field aliases for the fields of the inner select, and creating another `parseSqlRecord` method on the models. I think that as far as the query construction goes, the field alias generation with the table alias is the best I could come up with. It kind of mimics the behavior you need to follow when writing raw sql. I wish there was a way to combine the meat of the logic in the `parseSqlRecord` into one share function, but our normal `parseSqlRecord(record)` uses a Field as the lookup, and our `parseSqlRecord(record, alias)` method uses a string, which just looks at the column name. 

Any thoughts on this implementation would be appreciated! @timparsons @dmeidlin You might also be interested in checking this out since it is a base infrastructure change that will affect a lot of code in the future.

timparsons: I had some other comments written up on ideas for `parseSqlRecord` in particular, however, as I got looking more at this code, I think we are reaching the point of over-complicating a SQL query.  I think it's worth re-visiting the discussion of when to have a single query for all the data, and when to look at splitting it up into two calls.

IMO, I would split out the retrieval of a users programs and roles to a second SQL call.  I feel it will make the code cleaner, and easier to follow, as well as allowing the use of the original `parseSqlRecord` method.

ctucker3: Yep, I agree that in this case two calls is cleaner! In the future, I think we will still need to consider each complex query on a case by case basis to avoid slow calls and so that we don't end up doing manual joins for the database. I'm thinking of the phenotyping/genotyping data in particular when we are joining potentially millions of rows. 

Anyway, for this case, two calls doesn't add a lot of overhead so I think we are fine performance wise. I'll have the new code up soon. Thanks for the suggestion!

ctucker3: Changed and pushed. 

Also, updated the Program User endpoints and the api docs to return the program as well (they didn't before). To exclude the program from the endpoints, we will have to modify the way we are using jackson, I can describe a possible solution if anyway is interested. I had a conversation with Nick about this and one of the possible solutions we came up with is to set a standard that a response referencing a nested class should not have a higher level of detail than the nested classes own response. This forces a consistency with our responses. If anyone has further thoughts on this, I am open to discussing them!

docs/openapi.yaml
```diff
@@ -145,6 +180,7 @@ paths:
                     domain: "admin"
                   }
                   ]
+                  programRoles: []
```
nickpalladino, line 183: POST /users response example is missing the active field

ctucker3: changed and pushed

PR #28, opened 2020-06-30
INF-138 - Include data in response even if not populated

ctucker3:
One Line of code to fix it all, One line to find them,
One Line of code to bring them all and in the Jackson bind them

## Testing

- Tests run
- Data object is in the result of the response even if not results are returned.

### DIFF
--- a/src/main/java/org/breedinginsight/api/model/v1/response/DataResponse.java
+++ b/src/main/java/org/breedinginsight/api/model/v1/response/DataResponse.java
@@ -17,6 +17,7 @@
 
 package org.breedinginsight.api.model.v1.response;
 
+import com.fasterxml.jackson.annotation.JsonInclude;
 import lombok.*;
 import lombok.experimental.Accessors;
 
@@ -29,5 +30,6 @@ import java.util.List;
 @NoArgsConstructor
 @AllArgsConstructor
 public class DataResponse<T> {
+    @JsonInclude()
     private List<T> data;
 }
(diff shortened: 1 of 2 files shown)


src/main/java/org/breedinginsight/api/model/v1/response/DataResponse.java
```diff
@@ -29,5 +30,6 @@
 @NoArgsConstructor
 @AllArgsConstructor
 public class DataResponse<T> {
+    @JsonInclude()
```
nickpalladino, line 33: After looking up @JsonInclude, I see that it defaults to `JsonInclude.Include.ALWAYS`. For me personally seeing `JsonInclude.Include.ALWAYS` would make it more clear but IntelliJ doesn't like that and says it's redundant. Didn't know if anyone else had any thoughts on this? I'm ok with it either way just didn't know if we had a preferred style we should follow.

timparsons: I don't see a problem with leaving it as `@JsonInclude`

ctucker3: I was using `JsonInclude.Include.ALWAYS` at first in the other models, but I think I lean toward not including it now. Just having the `@JsonInclude` seems explicit enough to me. But, thats a good point on consistency, because some of our models will still have the `JsonInclude.Include.ALWAYS` so we should update those at some point.

ctucker3:
[INF-138 - API result for GET /programs not returning an empty data array](https://trello.com/c/UPReRYjh/239-inf-138-api-result-for-get-programs-not-returning-an-empty-data-array)

PR #30, opened 2020-07-24
Feature/ont 68

ctucker3:
Adds multi-error response to the trait upload endpoint. 

### Dependencies

To see the multiple errors on the front end, requires ONT-42. 

### Feature Changes

- Adds multi validation error of trait objects to the trait upload endpoint
- Adds multi parsing error response to the trait upload endpoint. 
- Checks for duplicates within the file. 
- Does not check from duplicates against the database -> these are silently ignored traits in POST /traits

### Design Choices

- Uses the same trait validation as POST /traits
- The trait parsing code allows all traits to be attempted to be parsed, and stops the upload process if there are any errors found. 
- If there is an uncaught parsing error, the request returns with that single response. The thought behind this was if we do not catch the parsing error, then it is unexpected, and therefore a critical error. 
- The multi parsing exception response returns the same error body structure as the validation error response for consistency of parsing on the front end. Our front end will need to adjust the row numbers if we want to show the excel file row.


nickpalladino:
[ONT-68 - Validate entire trait file](https://trello.com/c/WYTexsed/250-ont-68-validate-entire-trait-file)

src/main/java/org/breedinginsight/services/TraitUploadService.java
```diff
@@ -90,49 +95,46 @@ public ProgramUpload updateTraitUpload(UUID programId, CompletedFileUpload file,
...
             }
         } else if (mediaType.getName().equals(MediaType.XLS) ||
                    mediaType.getName().equals(MediaType.XLSX)) {
             try {
                 traits = parser.parseExcel(file.getInputStream());
             } catch(IOException | ParsingException e) {
                 log.error(e.getMessage());
-                throw new UnprocessableEntityException("Error parsing excel: " + e.getMessage());
+                throw new HttpBadRequestException("Error parsing excel: " + e.getMessage());
+            } catch (ValidatorException e){
```
nickpalladino, line 107: Since this is being rethrown and not allowed to flow past this any ValidatorException from the TraitFileParser will short circuit and not process other trait errors. For example if you have an invalid scale type and then other traits with say missing required fields, you'll only see the scale type error and not the others. Didn't we still want to check those other traits as well? I attached a file for an example.

[data_one_row_invalid_scale.xlsx](https://github.com/Breeding-Insight/bi-api/files/5007999/data_one_row_invalid_scale.xlsx)

src/test/java/org/breedinginsight/api/v1/controller/TraitUploadControllerIntegrationTest.java
```diff
@@ -78,11 +81,55 @@ public void setup() {
...
+        dsl.execute(fp.get("InsertTrait"));
+
         // Retrieve our new data
         validProgram = programDao.findAll().get(0);
 
     }
 
+    @Test
+    @Order(1)
+    void putTraitUploadDuplicatesInDB() {
```
nickpalladino, line 98: Do we have a card for being able to communicate back the rows that are duplicates for highlighting on the UI as specified in the design reference?

src/test/java/org/breedinginsight/api/v1/controller/TraitUploadControllerIntegrationTest.java
```diff
@@ -78,11 +81,55 @@ public void setup() {
...
+        dsl.execute(fp.get("InsertMethod"));
+        dsl.execute(fp.get("InsertScale"));
+        dsl.execute(fp.get("InsertTrait"));
+
         // Retrieve our new data
         validProgram = programDao.findAll().get(0);
 
     }
 
+    @Test
```
nickpalladino, line 96: Can we add a test for the case of an error in the TraitFileParser such as invalid scale and errors in other traits higher up in the service like missing required fields? Related to previous comment.

nickpalladino, line 96: I think we should also add a test case for the blank row case prior to data and row numbers in errors. It looks like there is an issue in the excel parser for this case that will need to be fixed.

[empty_then_2_rows.xlsx](https://github.com/Breeding-Insight/bi-api/files/5008188/empty_then_2_rows.xlsx)

src/main/java/org/breedinginsight/services/TraitUploadService.java
ctucker3: That was intentional because the type parsing errors may lead to false errors in the missing field validation check. For example, if a scale class was passed, but is a bad type, it will leave the scale class blank. If we let that keep going in the validation, it will throw a missing scale error, which would be wrong. 

So if the controller can't parse the file because of bad data types it will throw an error before moving onto other checks.

src/test/java/org/breedinginsight/api/v1/controller/TraitUploadControllerIntegrationTest.java
ctucker3: https://trello.com/c/cF12yK1j/81-ont-81-report-a-count-of-imported-new-and-duplicate-traits-to-the-ui

src/main/java/org/breedinginsight/services/TraitUploadService.java
nickpalladino: That makes sense for the particular trait that had the invalid scale, but couldn't you just drop that trait from further checks and check the other traits?

ctucker3: You could, but that trait may have errors such as missing trait name that wouldn't be included in the initial error return because it was ignored. It would then show up as an error when the user submitted the second time which may be confusing to the user because it didn't appear initially among the errors of similar type. 

Looking for @Breeding-Insight/dev-team feedback on this. Nick and I discussed this and didn't come to a conclusion. One thing we agreed on was that the ideal solution would require one large validation method (combination of all of the validations) that would require a lot of work and may not be worth it right now.

src/test/java/org/breedinginsight/api/v1/controller/TraitUploadControllerIntegrationTest.java
ctucker3: Added a test for this. I also fixed it in the parsing service. This will now pass and that empty row will be ignored.

ctucker3: Added test case

src/main/java/org/breedinginsight/services/parsers/trait/TraitFileParser.java
```diff
@@ -96,28 +107,38 @@
 
         Sheet excelSheet = convertCsvToExcel(records);
         List<ExcelRecord> excelRecords = ExcelParser.parse(excelSheet, TraitFileColumns.getColumns());
-        return excelRecordsToTraits(excelRecords);
 
+        try {
```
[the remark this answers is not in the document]
ctucker3: Changed and pushed

src/main/java/org/breedinginsight/services/parsers/trait/TraitFileParser.java
```diff
@@ -75,11 +80,17 @@
         }
 
         List<ExcelRecord> records = ExcelParser.parse(sheet, TraitFileColumns.getColumns());
-        return excelRecordsToTraits(records);
+
+        try {
```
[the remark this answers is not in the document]
ctucker3: Changed and pushed

src/main/java/org/breedinginsight/services/TraitUploadService.java
```diff
@@ -90,49 +95,46 @@ public ProgramUpload updateTraitUpload(UUID programId, CompletedFileUpload file,
                 traits = parser.parseCsv(file.getInputStream());
             } catch(IOException | ParsingException e) {
                 log.error(e.getMessage());
-                throw new UnprocessableEntityException("Error parsing csv: " + e.getMessage());
+                throw new HttpBadRequestException("Error parsing csv: " + e.getMessage());
             }
```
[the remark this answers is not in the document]
ctucker3: Removed the logging from the parseExcel call because it is logged in the controller method.

src/main/java/org/breedinginsight/services/TraitUploadService.java
ctucker3: Decided in team meeting to keep it how it is and implemented the ideal solution when we work on the trait editing page for trait imports.

ctucker3:
> Looks good other than api docs. Should put a note in there about the row indexing for trait-upload and example for 422 error shows a row index < 2.

Good catch! Updated and pushed.

eawoods:
I tried testing this with what's currently in `bi-web/develop` and with a (slightly modified) version of Deb's alfalfa traits file -- I can't get multi-errors to show, only one "Error parsing csv: Missing expected columns" banner at the top. What am I doing wrong?

nickpalladino:
> I tried testing this with what's currently in `bi-web/develop` and with a (slightly modified) version of Deb's alfalfa traits file -- I can't get multi-errors to show, only one "Error parsing csv: Missing expected columns" banner at the top. What am I doing wrong?

you need to run ONT-42 on bi-web to get the front end piece of that.

ctucker3:
> I tried testing this with what's currently in `bi-web/develop` and with a (slightly modified) version of Deb's alfalfa traits file -- I can't get multi-errors to show, only one "Error parsing csv: Missing expected columns" banner at the top. What am I doing wrong?

I can't see any obvious errors at a glance. I am going to return the missing columns in the error message. If we are confused by this error, a user will definitely be confused as well!

ctucker3:
> I tried testing this with what's currently in `bi-web/develop` and with a (slightly modified) version of Deb's alfalfa traits file -- I can't get multi-errors to show, only one "Error parsing csv: Missing expected columns" banner at the top. What am I doing wrong?

The error message for missing columns now returns what columns are missing. Also, like Nick was saying, make sure you use ONT-42 for bi-web for the multi-error display. I'll add that to this PR description.

PR #31, opened 2020-07-28
insert [NULL] into db when coordinates not specified

ctucker3:
The trello card explains it all, 

https://trello.com/c/03b8Z0jy/167-pro-77-bug-post-locations-controller-inserts-null-into-sql-column-when-there-is-no-data

- Made bi-api insert NULL into db when coordinates are not passed in. 
- Made bi-api not error out when reading NULL for coordinates from DB

### DIFF
--- a/src/main/java/org/breedinginsight/services/ProgramLocationService.java
+++ b/src/main/java/org/breedinginsight/services/ProgramLocationService.java
@@ -253,7 +253,7 @@ public class ProgramLocationService {
         placeEntity.setAccessibilityId(accessibilityId);
         placeEntity.setTopographyId(topographyId);
         placeEntity.setAbbreviation(programLocationRequest.getAbbreviation());
-        placeEntity.setCoordinates(JSONB.valueOf(coordinates));
+        placeEntity.setCoordinates(coordinates != null ? JSONB.valueOf(coordinates) : null);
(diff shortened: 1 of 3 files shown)


nickpalladino:
[PRO-77 - BUG - POST locations controller inserts 'null' into sql column when there is no data](https://trello.com/c/03b8Z0jy/254-pro-77-bug-post-locations-controller-inserts-null-into-sql-column-when-there-is-no-data)

src/main/java/org/breedinginsight/services/ProgramLocationService.java
```diff
@@ -208,7 +208,7 @@ public ProgramLocation create(AuthenticatedUser actingUser,
                 .accessibilityId(accessibilityId)
                 .topographyId(topographyId)
                 .abbreviation(programLocationRequest.getAbbreviation())
-                .coordinates(JSONB.valueOf(coordinates))
+                .coordinates(coordinates != null ? JSONB.valueOf(coordinates) : null)
```
nickpalladino, line 211: We should also do this for update as well right?

ctucker3: Ah yep, thanks for catching that! Changed and pushed. Test updated.

PR #32, opened 2020-07-31
Feature/pro 78

ctucker3:
Adds an update orcid endpoint as a temporary fix for assigning an orcid to a user. 

## Dependencies
- Is paired on PRO-78 on bi-web

## Design 

- Added an additional endpoint instead of modifying the existing endpoints so that the feature is easily removed and is less likely to cause errors when we remove it later.

### DIFF
--- a/src/main/java/org/breedinginsight/services/UserService.java
+++ b/src/main/java/org/breedinginsight/services/UserService.java
@@ -19,6 +19,7 @@ package org.breedinginsight.services;
 
 import lombok.extern.slf4j.Slf4j;
 import org.breedinginsight.api.auth.AuthenticatedUser;
+import org.breedinginsight.api.model.v1.request.OrcidRequest;
 import org.breedinginsight.api.model.v1.request.SystemRolesRequest;
 import org.breedinginsight.api.model.v1.request.UserRequest;
 import org.breedinginsight.dao.db.tables.daos.SystemRoleDao;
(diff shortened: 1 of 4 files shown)


src/main/java/org/breedinginsight/services/UserService.java
```diff
@@ -291,4 +292,27 @@ public User updateRoles(AuthenticatedUser actingUser, UUID userId, SystemRolesRe
...
+
+        BiUserEntity biUser = dao.fetchOneById(userId);
+
+        if (biUser == null) {
+            throw new DoesNotExistException("UUID for user does not exist");
+        }
+
+        List<BiUserEntity> biUserWithOrcidList = dao.fetchByOrcid(orcidRequest.getOrcid());
+        for (BiUserEntity biUserWithOrcid: biUserWithOrcidList){
+            if (!biUserWithOrcid.getId().toString().equals(userId.toString())){
```
nickpalladino, line 309: I think you could do equals on UUID if you wanted.

src/test/java/org/breedinginsight/api/v1/controller/UserControllerIntegrationTest.java
```diff
@@ -839,4 +839,93 @@ public void getUserInactiveProgramsNotReturned() {
...
+    public void putUserOrcidInUseBySelf() {
+
+        String orcid = "0000-0000-0000-0000";
+        JsonObject requestBody = new JsonObject();
+        requestBody.addProperty("orcid", orcid);
+
+        Flowable<HttpResponse<String>> call = client.exchange(
+                PUT("/users/" + otherTestUser.getId().toString() + "/orcid", requestBody.toString())
+                        .contentType(MediaType.APPLICATION_JSON)
+                        .cookie(new NettyCookie("phylo-token", "test-registered-user")), String.class
```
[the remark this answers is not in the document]
ctucker3: changed and pushed

src/main/java/org/breedinginsight/services/UserService.java
ctucker3: changed and pushed

timparsons:
[PRO-78 - Manual entry of ORCID iD](https://trello.com/c/T8gBhns4/256-pro-78-manual-entry-of-orcid-id)

ctucker3:
> When logged in as a breeder and adding a new user in the Prog. User table, the orcid id input field is present but no validation message appears when the id is omitted and the new user saved; the user is just created successfully without the id. The second acceptance criteria for Breeders states there should be a block in this case.

Changed and pushed

PR #33, opened 2020-08-06
Feature/inf 172

ctucker3:
Adds a github action that runs on pull request creation and edit. The action builds bi-api and runs the tests. Success is determined by the build and the tests running successfully. Failure occurs if any tests fail or bi-api fails to build. 

### Checks
- Does not run on every push to feature branch --> passed
- Checking that edits to PR details does not cause action to run --> passed
- Check that new commits made during PR cause action to rerun --> passed
- Checking that reviews on PR do not cause action to run --> passed
- Check that comments on PR do not cause action to run --> passed

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,3 +1,20 @@
+#
+#    See the NOTICE file distributed with this work for additional information
+#    regarding copyright ownership.
+#
+#    Licensed under the Apache License, Version 2.0 (the "License");
+#    you may not use this file except in compliance with the License.
+#    You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS,
+#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+#
+
 name: maven build
 
 on:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,39 @@
+name: maven build
+
+on:
+  pull_request:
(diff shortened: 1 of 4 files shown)


.github/workflows/build.yml
```diff
@@ -0,0 +1,39 @@
...
+        image: postgres:latest
+        env:
+          POSTGRES_USER: postgres
+          POSTGRES_PASSWORD: postgres
+          POSTGRES_DB: bidb
+        ports:
+          - 5432:5432
+        options: --health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5
+
+    steps:
```
nickpalladino, line 23: I don't know about this.

ctucker3:
[INF-172 - Configure GitHub Actions to automatically build bi-api](https://trello.com/c/KE4hA4XQ/259-inf-172-configure-github-actions-to-automatically-build-bi-api)

.github/workflows/build.yml
```diff
@@ -0,0 +1,56 @@
...
+      - uses: actions/checkout@v2
+      - name: Set up JDK 13
+        uses: actions/setup-java@v1.3.0
+        with:
+          java-version: 13
+
+      - name: Build with Maven
+        run: mvn -B flyway:migrate install --file pom.xml --settings settings.xml
```
nickpalladino, line 48: Did you test the case of a build error vs a test failure error and that the output is clear about what the issue is?

ctucker3: Yep, I added an intentional build error just now (and fixed it). You can see the different runs here, https://github.com/Breeding-Insight/bi-api/actions. The actions run summary shows all of the log output from the command, so we will be able to debug it the same way as if we had built it locally.

nickpalladino: I wonder if it would be beneficial to separate the build and test into separate actions? Not run the tests as part of the build action? Something to discuss but I'll approve this either way.

PR #34, opened 2020-08-17
Feature/inf 200

ctucker3:
Adds a unique constraint to the db for orcid. I did not need to add a unique check and tests to bi-api, because these were already implemented on PRO-78. This is based onto PRO-78, so that should be review and approved first. 

## Testing

Try out this query after applying your flyway migrations:

```
insert into bi_user (orcid, created_by, updated_by)
select '0000-0003-0437-8310', id, id from bi_user bu where name = 'system' 
```

That will give you a unique constraint error.

### DIFF
--- a/src/main/resources/db/migration/V0.5.14__set-orcid-unique.sql
+++ b/src/main/resources/db/migration/V0.5.14__set-orcid-unique.sql
@@ -1,0 +1,19 @@
+/*
+ * See the NOTICE file distributed with this work for additional information
+ * regarding copyright ownership.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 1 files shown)


ctucker3:
[INF-200 - Make the orcid id column for the bi_user table unique](https://trello.com/c/jfg7HCvd/269-inf-200-make-the-orcid-id-column-for-the-biuser-table-unique)

ctucker3:
> Prevents repeat ORCIDs as described. It's a little awkward that a user can be created without ORCID -- I used the front end for testing as well, and could create a user with dupe ORCID, with error message that the user was entered without any ORCID. I realize this is a workaround, so it may not matter in the long run.
> 
> Fun fact: if you have duplicate ORCIDs in your database when starting `bi-api` using docker-compose, it will fail with messages about not being able to run flyway:migrate. Figured out how to fix that one on my own; error is unique to those who directly mess around with their testing db and should not need to be addressed in code. :-)

Agreed that the scenario is described is clunky. With the user creation and orcid assignment being two separate calls, there isn't too much we can do about that scenario. That will go away once we implement the real registration flow.

PR #35, opened 2020-08-17
prevent user from deactivating themselves

ctucker3:
Adds a check to not allow the user to deactivate their own account. 

## Dependencies

Is paired with ISSUE-5 PR in bi-web.

### DIFF
--- a/src/main/java/org/breedinginsight/api/v1/controller/UserController.java
+++ b/src/main/java/org/breedinginsight/api/v1/controller/UserController.java
@@ -157,11 +157,15 @@ public class UserController {
     public HttpResponse archiveUser(@PathVariable UUID userId){
 
         try {
-            userService.archive(userId);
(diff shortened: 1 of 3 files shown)


ctucker3:
[ISSUE-5 - A system administrator is able to deactivate themselves](https://trello.com/c/rd1m1V96/267-issue-5-a-system-administrator-is-able-to-deactivate-themselves)

ctucker3:
> Looks like there were a couple tests that have errors from older stuff (not part of this work). Guess I don't need to check this anymore because of the github actions but since I did run them, there you go. Actually just saw checks are passing for actions so not sure what's going on.

Can you try to do a `mvn clean install` on your end and retry? Sometimes with my intellij to run all tests I have to use `mvn test` instead of running with junit. The build is successful and shows all the tests as passing in github actions still after bringing up to date with develop.

PR #39, opened 2020-08-28
update brapi server reference

ctucker3:
Updates docker-compose to reference the new brapi-server image.


ctucker3:
[INF-205 - Change brapi-server to support text instead of varchar(255)](https://trello.com/c/UhYjCi4x/284-inf-205-change-brapi-server-to-support-text-instead-of-varchar255)

PR #41, opened 2020-09-14
[BI-603] - Traits: Unable to import a valid csv traits file

ctucker3:
Fixes issue with upload csv files on windows machines. The solution uses the Apache Tika library to check file extension and the file's magic numbers to determine the file type. Included tests to check that file types are parsed correctly. Included a test to make sure malicious users can't mask unsupported file types as supported file types. 

## Noteworthy Notes

I needed to cherry pick the fix for server request pollution that was merged into develop from ISSUE-9 so that all the tests would pass in github actions. This might require fixing a merge conflict when we merge back into develop, but it should be easy to fix. 

## Testing

Test an `xlsx`, `xls`, `csv` file and confirm they work. If you want to get spicy, take a small `exe` file or some other file type from your downloads, rename the extension to `csv` and give it a try. If you get an exceeded file size error, the file is too big, try something smaller. 

## Error states

Error states are the same as before. See unsupported mimetype error screenshot below:

screencapture-192-168-10-72-8080-programs-65bd16db-63a3-4c9c-a4bd-5fef00e1dd71-traits-import-2020-09-14-15_05_58


src/main/java/org/breedinginsight/services/TraitUploadService.java
```diff
@@ -86,20 +90,24 @@ public ProgramUpload updateTraitUpload(UUID programId, CompletedFileUpload file,
             throw new AuthorizationException("User not in program");
         }
 
-        Optional<io.micronaut.http.MediaType> type = file.getContentType();
-        io.micronaut.http.MediaType mediaType = type.orElseThrow(() -> new UnsupportedTypeException("File upload must have a mime type"));
+        MediaType mediaType;
+        try {
+            mediaType = mimeTypeParser.getMimeType(file);
+        } catch (IOException e){
+            throw new HttpBadRequestException("Could not read file");
```
nickpalladino, line 97: May be a little more clear to have this as "Could not determine file type" so as to not get confused with file read errors below.

nickpalladino (commented):
Glad Tika appears to be working out. Could you rename the PR title to BI-603 - Traits: Unable to import a valid csv traits file? If you want something else, let me know but that's what I've been following and I'd like things to be organized for the automated release notes.

ctucker3:
> Glad Tika appears to be working out. Could you rename the PR title to BI-603 - Traits: Unable to import a valid csv traits file? If you want something else, let me know but that's what I've been following and I'd like things to be organized for the automated release notes.

src/main/java/org/breedinginsight/services/TraitUploadService.java
ctucker3: Agreed, changed and pushed

nickpalladino:
> > Glad Tika appears to be working out. Could you rename the PR title to BI-603 - Traits: Unable to import a valid csv traits file? If you want something else, let me know but that's what I've been following and I'd like things to be organized for the automated release notes.
> 
> Oh yep, thanks for catching that! I've been doing `[BI-603] - Traits: Unable to import a valid csv traits file` because I like how easy it is to spot the issue number. Thoughts on doing it that way? Ultimately, it doesn't matter to me and I agree that it will be good to have a consistent naming scheme.

Ya, that works. The main thing is we have the id and title because by default github doesn't do that.

[BI-603]: https://breedinginsight.atlassian.net/browse/BI-603

nickpalladino (approved):

dmeidlin (approved):

PR #44, opened 2020-10-01
[BI-611] - Sort/Pagination/Filter

ctucker3:
Adds search/filter/pagination to the backend, specifically for programs and program locations. My plan is to implement search and sort for fields we can currently create through the UI. For these two models I implemented a few more fields for testing purposes. 

## Design Choice

The search/filter/paginate was added at the time of the response processing, after all the data had been retrieved. The other alternatives were to add limits, skips and where clauses to the sql query to filter down the data. This would only be part of the solution though because we would need to get the other part of our data from the brapi services. We would have to query all of the data from the brapi service because the brapi search does not have a `.contains` operator. We would then have to sort and filter the data again after the data came back, which would be nearly identical the methods in this solution. So, it would have been double the work (or more), for minimal performance gain. 

This solution also has the advantage that it works at the controller level, so it will be able to configure this feature on and off. This may come in handy when we implement a more advanced search/filter such as the elastic search flow Tim has been looking into. 

The `GET /programs` endpoint will have pagination and sorting
The `POST /programs/search` endpoint will have pagination/sorting/filtering

Instead of moving the pagination and sorting into the body from the query params into the body for the search endpoint, I kept them in the query. From what I was reading, this doesn't conflict with the specifications of API design. 

## Query Behavior

Our pagination is 1 indexed. Below I outline some queries of note and their behavior:

Case: page = 0
`/programs?page=0` -> Status 400

Case pageSize = 0
`/programs?pageSize=0` -> Status 400

Case sortFeild does not exist:
`/programs?sortField=not_exist` -> Status 400

Case searchField does not exist:
`/programs` body {fields: [ {field: not_exist, value: blah } ] } -> Status 400

## Code design

The sort/filter/paginate methods are generic, so it will be minimal work to get them up and running on new endpoints. For the endpoints you want to add processing to, you would create a `QueryMapper` where you map the query fields to the model fields. The QueryMapper returns a getter (specified as a lambda function in the method). 

## How it plays with bi-web

This breaks bi-web because bi-web sends back a request:

`programs?page=0&pageSize=0&showAll=true`

the 0 for page and pageSize will return a page response. a quick fix to this would be to go into the daos and where it has a call:

`api.call({ url: `${process.env.VUE_APP_BI_API_V1_PATH}/programs`, method: 'get', params: paginationQuery})`

change that to:

`api.call({ url: `${process.env.VUE_APP_BI_API_V1_PATH}/programs`, method: 'get'})`

It seems to work, and the mock pagination and sorting on the frontend will do its work as normal. Best to test this out when doing the work though. 

## Future Work Ideas

In the future it would be cool to try create a Factory to generate these query mappers and then annotate fields that we want included from the pojo. That way we could reuse nested pojos and not duplicate code (Say if species is annotated and is reference from the program pojo). Not sure if it would work, but would be cool to explore. 

Probably too much to explore and add this now though.


nickpalladino:
Just reading through the pull request description How it plays with bi-web section and it says it breaks it. Are we not going to be merging this pull request then until that's ready to be addressed?

ctucker3:
> Just reading through the pull request description How it plays with bi-web section and it says it breaks it. Are we not going to be merging this pull request then until that's ready to be addressed?

I think it would be best to implement the bi-api fix. It should be quick and I will likely have time to do it tomorrow. @timparsons might have more of an idea if we should fit that in the scope of this or not. I won't merge this PR until we decide how we want to move forward.

timparsons (approved):
Looks great! 

One comment, but nothing that needs to hold up this PR:

>The POST /programs/search endpoint will have pagination/sorting/filtering
Instead of moving the pagination and sorting into the body from the query params into the body for the search endpoint, I kept them in the query. From what I was reading, this doesn't conflict with the specifications of API design.

Only thought I had on this was if it'd be easier to be more consistent for the POST to have everything in the JSON body.  By doing that, there wouldn't be a mix of reading the POST's body and query parameters.  I think it comes down to preference more than anything (of which I personally have none).

timparsons:
> > Just reading through the pull request description How it plays with bi-web section and it says it breaks it. Are we not going to be merging this pull request then until that's ready to be addressed?
> 
> I think it would be best to implement the bi-api fix. It should be quick and I will likely have time to do it tomorrow. @timparsons might have more of an idea if we should fit that in the scope of this or not. I won't merge this PR until we decide how we want to move forward.

If the easier fix is in bi-api for the short term (and I'm assuming short term is until we implement proper pagination requests on bi-web), then I say go for it!

nickpalladino (approved):

ctucker3:
> If the easier fix is in bi-api for the short term (and I'm assuming short term is until we implement proper pagination requests on bi-web), then I say go for it!

Fixes have been implemented on bi-web. I will merge this once that PR is approved

https://github.com/Breeding-Insight/bi-web/pull/45

PR #46, opened 2020-10-06
[BI-613, BI-614, BI-616] - Sort/Pagination/Filter for traits, users, program users

ctucker3:
Also adds new sorting functionality. Can now sort a list of lists of strings (List<List<String>>). This is needed for our tables that list multiple values in a column, such as the programs in the users table, or roles (when/if we allow multiple roles on front end). 

Docs and tests have been updated.


src/main/java/org/breedinginsight/utilities/response/mappers/ProgramLocationQueryMapper.java
```diff
@@ -0,0 +1,62 @@
...
+                Map.entry("slope", ProgramLocation::getSlope),
+                Map.entry("createdAt", ProgramLocation::getCreatedAt),
+                Map.entry("updatedAt", ProgramLocation::getUpdatedAt),
+                Map.entry("createdByUserId",
+                        (location) -> location.getCreatedByUser() != null ? location.getCreatedByUser().getId() : null),
+                Map.entry("createdByUserName",
+                        (location) -> location.getCreatedByUser() != null ? location.getCreatedByUser().getName() : null),
+                Map.entry("updatedByUserId",
+                        (location) -> location.getUpdatedByUser() != null ? location.getUpdatedByUser().getId() : null),
+                Map.entry("updatedByUserName",
```
timparsons, line 47: I probably should have put this comment on #44 as well, but, it may be worth changing these strings to constants in this class, or maybe even a nested enum.  Same thing goes for the other `QueryMapper`s

timparsons (changes requested):
Comment left in regards to `QueryMapper`s.  Outside of that, looks good!

nickpalladino (approved):

src/main/java/org/breedinginsight/utilities/response/mappers/ProgramLocationQueryMapper.java
ctucker3: I was thinking since these are only used in this one place it wouldn't add any benefit defining those strings as constants. Since the incoming request strings get mapped to the getters, the map keys aren't referenced any where in the code. So normally I agree with the constants are enums pattern, but I was thinking it wasn't necessary here. What are your thoughts on it?

timparsons: Ok, good point.  Something to keep in mind if there's ever a direct use of the mapped fields in the future.

timparsons (approved):

PR #49, opened 2020-10-06
[BI-617] - Sort/paginate/filter for trait upload

ctucker3:
Updated trait upload services to parse the JSONB data from the database before returning to controller. For the sorting/filtering/pagination we needed something we could read and act on, so that seemed like the most logically course. It doesn't change the way the data is returned. 

There is one more point of failure in the upload process though. If for some reason the data was not stored properly during the file upload, there will be a failure when parsing this. But, that should never happen since the JSONB being stored in the database is rendered from Trait objects to begin with.

### DIFF
--- a/src/main/java/org/breedinginsight/model/ProgramUpload.java
+++ b/src/main/java/org/breedinginsight/model/ProgramUpload.java
@@ -30,6 +30,8 @@ import lombok.experimental.SuperBuilder;
 import org.breedinginsight.dao.db.tables.pojos.BatchUploadEntity;
 import org.jooq.Record;
 
+import java.util.List;
+
 import static org.breedinginsight.dao.db.Tables.BATCH_UPLOAD;
 
 @Getter
@@ -39,17 +41,22 @@ import static org.breedinginsight.dao.db.Tables.BATCH_UPLOAD;
 @SuperBuilder
 @NoArgsConstructor
 @JsonIgnoreProperties(value = { "createdBy", "updatedBy", "programId", "userId", "id"})
-public class ProgramUpload extends BatchUploadEntity {
+public class ProgramUpload<T> extends BatchUploadEntity {
 
     private Program program;
     private User user;
     private User createdByUser;
     private User updatedByUser;
+    private List<T> parsedData;
 
     // JSONB not working with jackson
     @JsonProperty("data")
     @JsonInclude(JsonInclude.Include.ALWAYS)
     public Trait[] getDataJson() throws JsonProcessingException {
+        if (parsedData != null) {
+            return (Trait[]) parsedData.toArray();
+        }
+
         ObjectMapper objMapper = new ObjectMapper();
         return objMapper.readValue(super.getData().data(), Trait[].class);
     }
--- a/src/main/java/org/breedinginsight/model/ProgramUpload.java
+++ b/src/main/java/org/breedinginsight/model/ProgramUpload.java
@@ -40,6 +40,7 @@ import static org.breedinginsight.dao.db.Tables.BATCH_UPLOAD;
 @Accessors(chain=true)
 @ToString
 @NoArgsConstructor
+@SuperBuilder(builderMethodName = "uploadBuilder")
 @JsonIgnoreProperties(value = { "createdBy", "updatedBy", "programId", "userId", "id"})
 public class ProgramUpload<T> extends BatchUploadEntity {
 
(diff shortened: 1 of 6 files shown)


src/main/java/org/breedinginsight/model/ProgramUpload.java
```diff
@@ -30,25 +31,27 @@
...
 
+import java.util.List;
+
 import static org.breedinginsight.dao.db.Tables.BATCH_UPLOAD;
 
 @Getter
 @Setter
 @Accessors(chain=true)
 @ToString
-@SuperBuilder
```
timparsons, line 39: Guessing this was removed because there was an ambiguous "builder()" method?  If so, the name of the builder for this class can be changed by updating the `@Superbuilder` annotation to something like: `@SuperBuilder(builderMethodName = "uploadBuilder")`

The use of the builder will need to be typed given that `ProgramUpload` now has generic support for the `parsedData` field.  An example of how to do this would be: `ProgramUpload.<T>uploadBuilder()...`

src/main/java/org/breedinginsight/utilities/response/ResponseUtils.java
```diff
@@ -0,0 +1,214 @@
...
+    public static <T> HttpResponse<Response<ProgramUpload>> getQueryResponse(
+            ProgramUpload upload, AbstractQueryMapper mapper, SearchRequest searchRequest, QueryParams queryParams) {
+        return processSearchResponse(upload, searchRequest, queryParams, mapper, new Metadata());
+    }
+
+    // Pagination and sort only
+    public static <T> HttpResponse<Response<ProgramUpload>> getQueryResponse(
+            ProgramUpload upload, AbstractQueryMapper mapper, QueryParams queryParams) {
+        return processSearchResponse(upload, null, queryParams, mapper, new Metadata());
+    }
```
timparsons, line 69: I think it'd be helpful to have the names of these methods reflect that they are for uploads instead of overloading the method name of `getQueryResponse`

src/main/java/org/breedinginsight/utilities/response/ResponseUtils.java
```diff
@@ -0,0 +1,214 @@
...
+        if (searchRequest != null){
+            data = search(data, searchRequest, mapper);
+        }
+        data = sort(data, queryParams, mapper);
+        Pair<List, Pagination> paginationResult = paginateData(data, queryParams);
+        metadata = constructMetadata(metadata, paginationResult.getRight());
+        return HttpResponse.ok(new Response(metadata, new DataResponse(paginationResult.getLeft())));
+    }
+
+    private static <T> HttpResponse<Response<ProgramUpload>> processSearchResponse(
```
timparsons, line 88: Same thing goes here for the method name

timparsons (changes requested):

nickpalladino (approved):

src/main/java/org/breedinginsight/utilities/response/ResponseUtils.java
ctucker3: How so? I plan on changing the names, but still wondering your thoughts as to the benefits of changing the method name instead of overloading?

src/main/java/org/breedinginsight/model/ProgramUpload.java
ctucker3: Perfect! Yeah the type ProgramUpload was messing things up and not liking `ProgramUpload.<Trait>builder()...`. I'll try out the named builder and see how that goes.

ctucker3: Changed and pushed

src/main/java/org/breedinginsight/utilities/response/ResponseUtils.java
ctucker3: Changed and pushed

timparsons: My main reason for saying that is because of this `HttpResponse<Response<ProgramUpload>>` being specifically typed to `ProgramUpload`, and the overloaded method has it as generice `<T>`

ctucker3: Created new methods for ProgramUpload and removed the generic typing from those methods.

timparsons (approved):

PR #50, opened 2020-10-15
[BI-280] - Add brapi program create and update on program create and update

ctucker3:
Creates brapi program on program create. Updates brapi program on program update. 

When the program is created, it is posted to the default brapi service. When we add the functionality to allow the program to sync with their own brapi service, the program and all other existing data will need to be created in that service. 

I did some rework on the program integration tests because the brapi integration broke a lot of them. Some general clean up is also included because those tests are a pain to use and debug.

### DIFF
--- a/src/main/java/org/breedinginsight/api/v1/controller/brapi/BrAPIServiceFilter.java
+++ b/src/main/java/org/breedinginsight/api/v1/controller/brapi/BrAPIServiceFilter.java
@@ -40,7 +40,7 @@ import javax.inject.Inject;
 import java.util.Map;
 import java.util.UUID;
 
-@Filter("/**/programs/**")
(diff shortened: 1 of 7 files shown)


nickpalladino (approved):
Discussed testing of what's sent to brapi services, not currently in automated tests. Is this something Deb would be willing to test manually? One concern we have with adding to automated tests is additional overhead. Currently only checked in code review.

DebWeigand:
> Discussed testing of what's sent to brapi services, not currently in automated tests. Is this something @DebWeigand would be willing to test manually? One concern we have with adding to automated tests is additional overhead. Currently only checked in code review.

Hey @nickpalladino - sorry for my late response.  I only recently saw this...  Yes, I'm happy to test this.  Perhaps we can get together soon and you can show me how I can do this.

dmeidlin (approved):

PR #54, opened 2020-10-22
[BI-379] Display web and API build version on UI

timparsons:
# Overview
Display the current version of both bi-web and bi-api on the page, and provide a link to the GitHub release (or commit if running `develop`).  This can be useful for debugging purposes, as well as for identifying what features are deployed to a given server.  The implementation of updating and display version info of bi-web is in https://github.com/Breeding-Insight/bi-web/pull/52.  Here's a design of placement of the version info:

## Implementation
The version information for bi-api is stored in `src/main/resources/version.properties`.  The version name is stored in the `version` property, and a link to the release (or commit) is stored in a custom property named `versionInfo`.  These properties are automatically updated and committed to the repo via GitHub Actions.  

When merging to `master`, the action will use the output from release-drafter to set `version` to the name generated by release-drafter and `versionInfo` to a link to the eventual tag that will be created for the release.

When merging to `develop`, a different action will update the `version` property to be whatever the current version number is, and then append the build number of the action.  For example, if the version name is `v0.2.0`, and the action is running for the 10th time, then the resulting version name on the `develop` branch will be `v0.2.0+10`.  The `versionInfo` property will be set to the merge commit that triggered the action.


src/main/java/org/breedinginsight/api/v1/controller/ServerInfoController.java
```diff
@@ -0,0 +1,49 @@
...
+import io.micronaut.security.rules.SecurityRule;
+import lombok.extern.slf4j.Slf4j;
+import org.breedinginsight.api.model.v1.response.ServerInfo;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Properties;
+
+@Slf4j
+@Controller("/${micronaut.bi.api.version}/serverInfo")
```
nickpalladino, line 33: We've been using kebab case for url paths. May be more clear to have the endpoint specified in the Get annotation, at least that's what I'm used to looking at. Also could you add this endpoint to the api docs?

src/test/java/org/breedinginsight/api/v1/controller/ServerInfoControllerIntegrationTest.java
```diff
@@ -0,0 +1,57 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+@MicronautTest
+public class ServerInfoControllerIntegrationTest {
+    @Inject
+    @Client("/${micronaut.bi.api.version}")
+    private RxHttpClient client;
+
+    @Test
+    public void getVersionInfo() throws IOException {
```
nickpalladino, line 44: Could possibly have another test where the version info can't be loaded, should give a 500 based on the code. Is that what we want, not sure what kind of diagnostic information is given in that case?

nickpalladino (changes requested):

src/main/java/org/breedinginsight/api/v1/controller/ServerInfoController.java
timparsons: Changed and pushed

src/test/java/org/breedinginsight/api/v1/controller/ServerInfoControllerIntegrationTest.java
timparsons: My personal opinion is that it's a little overkill to have a negative test for this.  If others feel strongly that there should be one, then I'll add it.

nickpalladino (approved):

dmeidlin (approved):

PR #58, opened 2020-11-16
[BI-601] - Program security

ctucker3:
Implements program security on the backend. For most endpoints, the security is checked at the controller level with the `@ProgramSecured` annotation. The one exception is `getPrograms` which gets the users enrolled programs during the call to return only the programs they are enrolled in. 

I also added a query to get the user at the beginning of every call. It will only be queried once per call. This was being done before this feature, so we won't notice a performance effect. 

The `@ProgramSecured` annotation can either accept roles directly, or role groups which are groups of roles that we can define for ease of use. 

I went through and annotated all of our existing controllers except for the specific locations ones like EnvironmentTypeController which we don't use right now.


src/test/java/org/breedinginsight/api/auth/rules/ProgramSecuredAnnotationRuleIntegrationTest.java
```diff
@@ -0,0 +1,214 @@
...
+
+        HttpClientResponseException e = Assertions.assertThrows(HttpClientResponseException.class, () -> {
+            HttpResponse<String> response = call.blockingFirst();
+        });
+
+        assertEquals(HttpStatus.FORBIDDEN, e.getStatus());
+    }
+
+    @Test
+    @Order(12)
```
nickpalladino, line 178: interesting (talking about the @Order(12) haha)

src/main/java/org/breedinginsight/api/auth/SecurityService.java
```diff
@@ -21,29 +21,94 @@
...
+        this.programDao = programDao;
+        this.userDAO = userDAO;
+        this.actingUserProvider = actingUserProvider;
+    }
+
     public AuthenticatedUser getUser() {
+
+        // If our acting user has already been set, don't query them again
+        if (actingUserProvider.get().getActingUser() != null) {
+            return actingUserProvider.get().getActingUser();
```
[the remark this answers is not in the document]
ctucker3: This is only for that particular user. The actingUserProvider is `@RequestScope` so it only belongs to that request.

src/main/java/org/breedinginsight/api/auth/rules/ProgramSecuredAnnotationRule.java
```diff
@@ -51,18 +60,35 @@ public SecurityRuleResult check(HttpRequest request, @Nullable RouteMatch routeM
 
             String programId = (String) routeMatch.getVariableValues()
                     .get("programId");
+
             if (methodRoute.hasAnnotation(ProgramSecured.class)) {
+                if (programId == null) {
+                    throw new HttpServerException("Endpoint does not have program id to check roles against");
+                }
+
+                if (!programDAO.existsById(UUID.fromString(programId))) {
```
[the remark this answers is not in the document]
ctucker3: Good question, I hadn't thought of that. I would say we should leave the active/not active checking in the controller if it is necessary. Not sure if we will have endpoints that deal with inactive programs, like a reactivate endpoint, so I want to account for the possibility of that.

src/test/java/org/breedinginsight/api/auth/rules/ProgramSecuredAnnotationRuleIntegrationTest.java
ctucker3: Hahah copy and paste strikes again! Changed and pushed

nickpalladino (approved):

src/main/java/org/breedinginsight/daos/ProgramUserDAO.java
```diff
@@ -115,6 +115,7 @@ public ProgramUser getProgramUser(UUID programId, UUID userId) {
         Result<Record> records = getProgramUsersQuery(createdByUser, updatedByUser)
                 .where(PROGRAM_USER_ROLE.USER_ID.eq(userId))
                 .and(PROGRAM.ACTIVE.eq(true))
+                .and(PROGRAM_USER_ROLE.ACTIVE.eq(true))
```
timparsons, line 118: I'd suggest renaming this method to `getActiveProgramUser` (and making an additional method to fetch all users) so that it is clear that a user could exist, but be inactive, and therefore nothing would be returned by this method.

timparsons (changes requested):
One change to a method name to bring clarity to what it's doing, but other than that, looks great!

src/main/java/org/breedinginsight/daos/ProgramUserDAO.java
ctucker3: I removed that new line, so it returns all program users now. This was a quick addition I thought was the fix, but it was not.

timparsons (approved):

PR #60, opened 2020-11-30
[BI-85,BI-646] - POST traits from UI

ctucker3:
Some fixes to allow POST traits to work directly from the UI. 

Added an optional deserializer for List<String> and String[] to remove blank strings and nulls from incoming lists. I couldn't figure out how to get it to happen for every List<String>, so I stuck with this method. There is a deserializer for all incoming strings that trims them and replaces them with null if they are blank. 

The trait upload now saves traits by POSTing a confirmation to the trait upload controller, which gets the json from the db and saves that. I could have done this without requiring the upload ID, but having the upload ID makes it more stateless because you are acting on a specific resource. 

Mixes in a bit of BI-686 in this one. I was working on these at the same time so there is a bit of overlap.

### DIFF
--- a/src/main/java/org/breedinginsight/api/v1/controller/ProgramController.java
+++ b/src/main/java/org/breedinginsight/api/v1/controller/ProgramController.java
@@ -405,7 +405,7 @@ public class ProgramController {
         }
     }
 
-    @Get("/programs/{programId}/observation_level")
+    @Get("/programs/{programId}/observation-level")
     @Produces(MediaType.APPLICATION_JSON)
     @ProgramSecured(roleGroups = {ProgramSecuredRoleGroup.ALL})
     public HttpResponse<Response<DataResponse<ProgramObservationLevel>>> getProgramObservationLevels(@PathVariable UUID programId)
--- a/src/main/java/org/breedinginsight/api/v1/controller/ProgramController.java
+++ b/src/main/java/org/breedinginsight/api/v1/controller/ProgramController.java
@@ -28,17 +28,20 @@ import org.breedinginsight.api.auth.*;
 import org.breedinginsight.api.model.v1.request.*;
 import org.breedinginsight.api.model.v1.request.query.QueryParams;
 import org.breedinginsight.api.model.v1.request.query.SearchRequest;
+import org.breedinginsight.api.model.v1.response.metadata.Metadata;
+import org.breedinginsight.api.model.v1.response.metadata.Pagination;
+import org.breedinginsight.api.model.v1.response.metadata.Status;
+import org.breedinginsight.api.model.v1.response.metadata.StatusCode;
 import org.breedinginsight.api.model.v1.validators.QueryValid;
 import org.breedinginsight.api.model.v1.validators.SearchValid;
+import org.breedinginsight.model.*;
+import org.breedinginsight.services.ProgramObservationLevelService;
 import org.breedinginsight.services.exceptions.*;
 import org.breedinginsight.utilities.response.mappers.ProgramLocationQueryMapper;
 import org.breedinginsight.utilities.response.mappers.ProgramQueryMapper;
 import org.breedinginsight.api.model.v1.response.DataResponse;
 import org.breedinginsight.api.model.v1.response.Response;
 import org.breedinginsight.api.v1.controller.metadata.AddMetadata;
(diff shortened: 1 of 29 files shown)


src/test/java/org/breedinginsight/api/v1/controller/TraitUploadControllerIntegrationTest.java
```diff
@@ -105,6 +115,9 @@ public void setup() {
 
         // Retrieve our new data
         validProgram = programDao.findAll().get(0);
+
```
nickpalladino, line 118: Any reason this was left in here commented out?

nickpalladino (changes requested):

src/main/java/org/breedinginsight/api/v1/controller/TraitUploadController.java
```diff
@@ -143,5 +143,23 @@ public HttpResponse deleteTraitUpload(@PathVariable UUID programId) {
 
     }
 
+    @Post("/programs/{programId}/trait-upload/{traitUploadId}")
```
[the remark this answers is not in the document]
ctucker3: Added and pushed

src/main/java/org/breedinginsight/api/v1/controller/ProgramController.java
```diff
@@ -399,4 +405,20 @@ public HttpResponse archiveProgramLocation(@PathVariable UUID programId,
         }
     }
 
+    @Get("/programs/{programId}/observation_level")
```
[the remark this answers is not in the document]
ctucker3: changed and pushed

src/main/java/org/breedinginsight/api/v1/controller/ProgramController.java
```diff
@@ -399,4 +405,20 @@ public HttpResponse archiveProgramLocation(@PathVariable UUID programId,
...
     }
 
+    @Get("/programs/{programId}/observation-levels")
+    @Produces(MediaType.APPLICATION_JSON)
+    @ProgramSecured(roleGroups = {ProgramSecuredRoleGroup.ALL})
+    public HttpResponse<Response<DataResponse<ProgramObservationLevel>>> getProgramObservationLevels(@PathVariable UUID programId)
+            throws DoesNotExistException {
+        List<ProgramObservationLevel> programObservationLevels = programObservationLevelService.getByProgramId(programId);
+
+        List<org.breedinginsight.api.model.v1.response.metadata.Status> metadataStatus = new ArrayList<>();
```
timparsons, line 415: You can remove the package info, and just leave it as `Status`

src/test/java/org/breedinginsight/api/v1/controller/ProgramControllerIntegrationTest.java
```diff
@@ -2282,6 +2281,22 @@ public void searchProgramUsers() {
         assertEquals(1, pagination.get("currentPage").getAsInt(), "Wrong current page");
     }
 
+    @Test
+    public void getObservationLevels() {
+        Flowable<HttpResponse<String>> call = client.exchange(
+                GET("/programs/" + otherProgram.getId() + "/observation-level")
```
timparsons, line 2287: need to add an `s` at the end of the path -> `/observation-levels`

timparsons (changes requested):
Overall looks good! Only a couple of minor things to change.

src/test/java/org/breedinginsight/api/v1/controller/ProgramControllerIntegrationTest.java
ctucker3: changed and pushed

src/main/java/org/breedinginsight/api/v1/controller/ProgramController.java
ctucker3: changed and pushed

src/test/java/org/breedinginsight/api/v1/controller/TraitUploadControllerIntegrationTest.java
ctucker3: removed and pushed

timparsons (approved):

nickpalladino (approved):

PR #61, opened 2020-12-02
[BI-671,BI-679] - Create new trait - Scale class 'ordinal scale' and Scale class 'nominal class'

ctucker3:
Adds a check to check for missing or blank category labels and values. For nominal, only the value is checked. Both label and value are checked for ordinal.


timparsons (approved):

ctucker3:
@eawoods Can you review the delete buttons I put on the category rows? They are different from what you had in the design reference. These ones are shipped with Bulma so they saved me some time rather than creating those custom X boxes in the design reference. 

screencapture-localhost-8080-programs-876e20fc-55db-4d86-aee5-028dae831784-traits-list-2020-12-07-15_42_42

nickpalladino (approved):

PR #62, opened 2020-12-14
[BI-685] - Map backend trait validation errors to front end

ctucker3:
Changes unique trait check to be determined on by trait name. 

Updates to the validation errors so that you can indefinitely nest errors. This was needed to parse errors for specific categories. Here is an example of the new error format:

screencapture-jsonviewer-stack-hu-2020-12-14-13_33_37

### DIFF
--- a/src/main/java/org/breedinginsight/daos/TraitDAO.java
+++ b/src/main/java/org/breedinginsight/daos/TraitDAO.java
@@ -380,7 +380,7 @@ public class TraitDAO extends TraitDao {
 
             Result<Record> records = dsl.select()
                     .from(newTraits)
-                    .join(TRAIT).on(TRAIT.TRAIT_NAME.likeIgnoreCase(newTraits.field("new_trait_name")))
+                    .join(TRAIT).on(TRAIT.TRAIT_NAME.upper().equalIgnoreCase(newTraits.field("new_trait_name")))
                     .join(PROGRAM_ONTOLOGY).on(TRAIT.PROGRAM_ONTOLOGY_ID.eq(PROGRAM_ONTOLOGY.ID))
                     .join(PROGRAM).on(PROGRAM_ONTOLOGY.PROGRAM_ID.eq(PROGRAM.ID))
                     .join(SCALE).on(TRAIT.SCALE_ID.eq(SCALE.ID))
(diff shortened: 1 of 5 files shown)


src/main/java/org/breedinginsight/daos/TraitDAO.java
```diff
@@ -366,39 +366,28 @@ private Trait parseTraitRecord(Record record, BiUserTable createdByUser, BiUserT
...
                     .where(PROGRAM.ID.eq(programId))
                     .fetch();
 
-
             for (Record record: records) {
                 Trait trait = Trait.parseSqlRecord(record);
-                Scale scale = Scale.parseSqlRecord(record);
-                Method method = Method.parseSqlRecord(record);
-                trait.setScale(scale);
-                trait.setMethod(method);
```
timparsons, line 401: Why were these taken out?

src/main/java/org/breedinginsight/services/validators/TraitFileValidatorError.java
```diff
@@ -84,22 +84,27 @@ public ValidationError getMissingScaleCategoriesMsg() {
     }
 
     @Override
-    public ValidationError getBlankScaleCategoryLabelMsg(Integer categoryIndex) {
+    public ValidationError getBadScaleCategory() {
+        return new ValidationError("scale.categories", "Scale categories contain errors", HttpStatus.UNPROCESSABLE_ENTITY);
```
timparsons, line 88: I think "scale.categories" needs to change to "Scale Categories"

src/main/java/org/breedinginsight/services/validators/TraitValidatorService.java
```diff
@@ -195,11 +202,8 @@ public ValidationErrors checkDuplicateTraitsInFile(List<Trait> traits, TraitVali
         Map<String, List<Integer>> namesMap = new HashMap<>();
         for (Integer i = 0; i < traits.size(); i++) {
             Trait trait = traits.get(i);
-            if (trait.getTraitName() != null &&
-                    trait.getScale() != null && trait.getScale().getScaleName() != null
-            ){
-                String key = String.format("%s-%s", trait.getTraitName().toLowerCase(),
-                        trait.getScale().getScaleName().toLowerCase());
+            if (trait.getTraitName() != null){
+                String key = String.format("%s", trait.getTraitName().toLowerCase());
```
timparsons, line 206: The `String.format("%s"` is redundant, and can just be `String key = trait.getTraitName().toLowerCase();`

timparsons (changes requested):

src/main/java/org/breedinginsight/daos/TraitDAO.java
ctucker3: I took those out of the join to save some processing on the query. Looking at it again though, taking them out would be misleading for what the method name is. Thanks for pointing this out!

src/main/java/org/breedinginsight/services/validators/TraitFileValidatorError.java
ctucker3: Changed and pushed

src/main/java/org/breedinginsight/services/validators/TraitValidatorService.java
ctucker3: Changed and pushed

timparsons (approved):

nickpalladino (changes requested):
Could you add a test for category errors?

ctucker3:
> Could you add a test for category errors?

The trait upload controller contains tests. Tests added for trait controller.

nickpalladino (approved):

PR #63, opened 2020-12-14
[BI-692] - Automatically set scale class to `Numerical` with `Computation` method class

ctucker3:


src/main/java/org/breedinginsight/services/TraitService.java
```diff
@@ -269,4 +273,18 @@ public void assignTraitsProgramObservationLevel(List<Trait> traits, UUID program
...
+
+    public void preprocessTraits(List<Trait> traits) {
+
+        // Set data type to numerical when method class is computation
+        for (Trait trait: traits) {
+            if (trait.getMethod() != null && trait.getMethod().getMethodClass() != null &&
+                trait.getMethod().getMethodClass().equalsIgnoreCase(Method.COMPUTATION_TYPE)) {
+                if (trait.getScale() != null) {
+                    trait.getScale().setDataType(DataType.NUMERICAL);
+                }
```
timparsons, line 285: I thought we had agreed that the scale would always be set to `NUMERICAL`, even if there was a value previously set?

timparsons (changes requested):

src/main/java/org/breedinginsight/services/TraitService.java
ctucker3: It does do that. The only case where it wouldn't set the class to numerical is when no `Scale` is passed in, which would be caught by the checks after this section. If the user doesn't pass in a scale I want them to get an error that they didn't pass in a scale. If I'm not properly seeing what you are referencing, I'm open to a chat!

timparsons: Ahh! I misread this code yesterday.  Thanks!

timparsons (approved):

nickpalladino (changes requested):
Could you add a test case for this?

ctucker3:
> Could you add a test case for this?

Tests added

nickpalladino (approved):

PR #64, opened 2020-12-16
[BI-625] Configure TLS (HTTPS) for sandbox/qa environments

dmeidlin:
The scipt entrypoint.sh was changed so that a user may pass custom command line arguments to the java command that builds and runs the jar file. This makes it possible to enable remote debugging so that an IDE can be used to debug the api while running as part of teh entire bi-docker-stack.


nickpalladino (commented):
What's going on with all the source file changes, need to be rebased or somthing?

dmeidlin:
> What's going on with all the source file changes, need to be rebased or somthing?

It's not clear what happened. The branch was already rebased, but for some reason Github included the src file changes back to the original base commit when I made the PR.

PR #70, opened 2021-02-05
BI-674 - Program Definable BrAPI Service URL

nickpalladino:
## Changes
- Added brapiURL to program
- Added a defaultBrapiUrl config variable
  - Kept the geno, pheno, core but pointed all to default or program defined URL in case we want to set them individually in future
- Added numUsers to program (read-only generated on request, not stored)

## Notes
- There is an associated breedbase PR to add returning of HTTP status codes (should update to BI-674 in sgn repo)
  - Breedbase will error out on duplicate program names or program names longer than 20 characters
- Put client stuff for doing serverinfo in provider just to keep settings together (timeout, etc)
- Had to remove application-test.yml from .gitignore and add it to get tests to run without having dependency injection error


src/test/resources/application-test.yml
```diff
@@ -0,0 +1,53 @@
+#
```
timparsons, line 1: The only other thing I can think to try so that you wouldn't have to add this file to the repo would be to set a default value for `brapi.server.default-url` in `application.yml`

timparsons (approved):
Looks good!

src/main/java/org/breedinginsight/services/ProgramService.java
```diff
@@ -111,6 +128,7 @@ public Program create(ProgramRequest programRequest, AuthenticatedUser actingUse
                     .abbreviation(programRequest.getAbbreviation())
                     .objective(programRequest.getObjective())
                     .documentationUrl(programRequest.getDocumentationUrl())
+                    .brapiUrl(brapiUrl)
```
nickpalladino, line 131: Was talking with @DebWeigand and realized the default brapi URL should probably be saved to the database and not be left blank. If left blank, the .env default brapi URL could be changed and it would affect previously created programs.

DebWeigand (changes requested):
while testing and working with @nickpalladino , we discovered we are unable to enter the default docker hostname into the BrAPI field.  Receiving message "BrAPI URL specified is not supported".

src/main/java/org/breedinginsight/services/ProgramService.java
timparsons: I think we could address this later.  Maybe make a card for this?

nickpalladino: Talked about this, going to leave and also change the brapiUrl to return "System Default" instead of the config default URL on programs without a custom brapi URL.

src/test/resources/application-test.yml
nickpalladino: If you did that wouldn't you be limited to only changing the value in application.yml because it seems like it would still have trouble finding the env?

nickpalladino:
> while testing and working with @nickpalladino , we discovered we are unable to enter the default docker hostname into the BrAPI field. Receiving message "BrAPI URL specified is not supported".

Will no longer show the docker URL in the programs table with the change above and as discussed on slack should just not specify default to use the system default and not try to manually set it.

src/test/resources/application-test.yml
timparsons: No, you can always override a value in `application.yml` by specifying the same hierarchy in `application-prod/test/etc.yml`

nickpalladino: For github actions though it doesn't seem like you'd be able to otherwise it would be working now right?

src/main/java/org/breedinginsight/services/ProgramService.java
nickpalladino: Pushed changes

src/test/resources/application-test.yml
nickpalladino: Added default url to pom and removed application-test. Looks like it worked.

src/main/java/org/breedinginsight/daos/ProgramDAO.java
```diff
@@ -114,7 +131,13 @@ public ProgramDAO(Configuration config, DSLContext dsl, BrAPIProvider brAPIProvi
 
         // Parse the result
         for (Record record: queryResult){
+            if (record.getValue(PROGRAM.BRAPI_URL) == null) {
+                record.setValue(PROGRAM.BRAPI_URL, SYSTEM_DEFAULT);
+            }
```
ctucker3, line 136: Would it be better to store the default brapi url with the program? The scenario I'm thinking of is if we have default system A and create a Program 1, but then we switch our default system to system B, Program 1 will be looking in default system B for its data, but it will not be there. 

Edit: I see this was addressed below, but leaving my comment above in case it is talking about a different thing. What was the decision for this? To fix it now, or leave it for later?

src/main/java/org/breedinginsight/daos/ProgramDAO.java
```diff
@@ -124,20 +147,50 @@ public ProgramDAO(Configuration config, DSLContext dsl, BrAPIProvider brAPIProvi
...
+        String phenoUrl = defaultBrAPIPhenoUrl;
+
+        // only storing one program brapi url for now so set all to that one
+        if (!StringUtils.isBlank(programEntity.getBrapiUrl())) {
+            String brapiUrl = programEntity.getBrapiUrl();
+            coreUrl = brapiUrl;
+            genoUrl = brapiUrl;
+            phenoUrl = brapiUrl;
+        }
+
```
ctucker3, line 171: Looks like there is just one url for the program now. Should we reflect that in this function? Or are we keeping it this way in case we want to have multiple urls in the future? If we want to change it, maybe we can create a new card so it doesn't inflate this feature?

src/main/java/org/breedinginsight/daos/ProgramDAO.java
```diff
@@ -124,20 +147,50 @@ public ProgramDAO(Configuration config, DSLContext dsl, BrAPIProvider brAPIProvi
...
+        // for now just check for 200 response, in future we could check actual required endpoints
+        try {
+            ApiResponse<BrAPIServerInfoResponse> response = serverInfoAPI.serverinfoGet(BrAPIWSMIMEDataTypes.APPLICATION_JSON);
+        } catch (ApiException e) {
+            log.error(e.getMessage());
+            supported = false;
+        }
+        return supported;
+    }
```
ctucker3, line 192: This seems like more of a service method than a dao method. I'm cool with it staying this way if you prefer it in the dao though.

ctucker3 (changes requested):
Might not need any changes, but I have some questions.

src/main/java/org/breedinginsight/daos/ProgramDAO.java
nickpalladino: Ya, talked about this with Tim and decided to leave it as-is for now.

src/main/java/org/breedinginsight/daos/ProgramDAO.java
nickpalladino: Kept it because eventually we might want to have the ability to set these independently and not just one program URL.

src/main/java/org/breedinginsight/daos/ProgramDAO.java
nickpalladino: I think I put it there initially because we've been doing all the brapi API calls in the DAOs. I could see how there is some extra logic there around it that could be better in the service. If it's not a big deal I'll probably just leave it.

ctucker3 (approved):

DebWeigand (approved):

DebWeigand (approved):

PR #71, opened 2021-02-11
[BI-732,BI-729] - Archive traits

ctucker3:
Sets trait to active/inactive in our system. Sends `archived` or `active` through brapi's `variable.status` field. 

Need to add tests still and update the docs.


timparsons (approved):

timparsons (approved):
Code looks good.  Approving even though tests need to be added/updated

nickpalladino (approved):

PR #73, opened 2021-02-16
[BI-641] - Validate lower and upper limits for correctness

ctucker3:
Looks like a big PR, but is mostly changes to test import files. The change to BigDecimal in the scale class required a few extra changes. 

This is based off on BI-645, so that should be reviewed and merged before the review of this.


src/main/java/org/breedinginsight/daos/TraitDAO.java
```diff
@@ -359,9 +367,13 @@ public Trait updateTraitBrAPI(Trait trait, Program program) {
             existingVariable.getScale().setDataType(brApiTraitDataType);
             existingVariable.getScale().setDecimalPlaces(trait.getScale().getDecimalPlaces());
             BrAPIScaleValidValues brApiScaleValidValues = new BrAPIScaleValidValues()
-                    .categories(trait.getScale().getCategories())
-                    .max(trait.getScale().getValidValueMax())
-                    .min(trait.getScale().getValidValueMin());
+                    .categories(trait.getScale().getCategories());
+            if (trait.getScale().getValidValueMax() != null) {
```
timparsons, line 371: My suggestion would be to have the `BigDecimal#setScale` things done directly in the `Scale` class.  This would mean overriding the `setValidValueMax` and `setValidValueMin` methods that are currently handled by Lombok

timparsons (changes requested):

src/main/java/org/breedinginsight/daos/TraitDAO.java
nickpalladino: Then the max is less than valid value min check would be after rounding which I guess is what you'd want since that's what's being sent out.

ctucker3: Wouldn't that be the same as if we had kept the scale valid value max and min as an integer? 

I like the idea of having it in the class though. Maybe I could add a getBrAPIValidMax and min?

DebWeigand (changes requested):
The row numbers in the error messages are off by 2

minMax_2

src/main/java/org/breedinginsight/daos/TraitDAO.java
timparsons: Yeah, good question. Maybe I missed why that was changed to BigDecimal from Integer, @ctucker3 do you know?

ctucker3: Tim and I talked about this. The BigDecimal doesn't have a use since it will never be persisted. I think this card was initially created when we were going to bring all of the trait data into our schema. 

I reverted all of those related changes and just implemented the min/max check.

ctucker3:
> The row numbers in the error messages are off by 2
> 
> minMax_2

I checked your file with the new changes for this card and the errors started off at row 2. Let me know if it is working for you now as well.

timparsons (approved):

DebWeigand (approved):

PR #74, opened 2021-02-16
[BI-703] - Move hardcoded brapi page size from breedbase to bi-api

ctucker3:


src/main/java/org/breedinginsight/daos/TraitDAO.java
```diff
@@ -75,8 +75,10 @@ public TraitDAO(Configuration config, DSLContext dsl, BrAPIProvider brAPIProvide
         Map<UUID, Trait> dbVariablesMap = dbVariables.stream().collect(Collectors.toMap(Trait::getId, p -> p));
 
         // Get brapi variables
-        VariableQueryParams variablesRequest = new VariableQueryParams()
-                                                               .externalReferenceSource(referenceSource);
+        VariableQueryParams variablesRequest = new VariableQueryParams();
+        variablesRequest.externalReferenceSource(referenceSource);
+        variablesRequest.pageSize(10000);
```
nickpalladino, line 80: Should we make this configurable or shared in case we have to do it anywhere else before handling properly?

nickpalladino: just gonna leave for now until/if it's used elsewhere

nickpalladino (approved):

dmeidlin (approved):

PR #75, opened 2021-03-01
[BI-765] - Set up endpoint to GET import configs

ctucker3:
Endpoint is running as expected. There is another card to configure the germplasm import. We can add more imports as we need them. 

https://breedinginsight.atlassian.net/browse/BI-768

This implementation is driven by some reflection and mainly annotations in order to generate the import configs. The idea is that we can generate all of the base brapi objects like germplasm, cross, in a flattened form (no nesting) so then be used to generate the configs. That way our code is also our config. Each import type will handle the construction and POSTing of brapi objects in a `process()` method (more to come later). 

I am not writing tests yet. I'll do that once I have the whole flow finished (saving mapping, posting to brapi, etc). I'm doing this just so I can see the whole flow implementation, and then decide what needs to be tested. 

I am also going to do api docs later in case the json structure changes. This way I can minimize time restructuring json structure. 

Future Ideas: 
- Tim was going to look into a way that we could run some micronaut code on startup to generate the type config on startup so that we didn't need to do it on every call. 
- Right now I get all the imports using a `Reflections` library and looking for an annotation. There might be a way to do this with micronaut.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -350,6 +350,11 @@
             <version>4.3</version>
             <scope>compile</scope>
         </dependency>
+        <dependency>
+            <groupId>org.reflections</groupId>
+            <artifactId>reflections</artifactId>
+            <version>0.9.12</version>
+        </dependency>
     </dependencies>
     <build>
         <plugins>
--- a/src/main/java/org/breedinginsight/api/v1/controller/ImportController.java
+++ b/src/main/java/org/breedinginsight/api/v1/controller/ImportController.java
@@ -52,8 +52,6 @@ public class ImportController {
     @Secured(SecurityRule.IS_ANONYMOUS)
     @AddMetadata
     public HttpResponse<Response<DataResponse<ImportConfig>>> getImportTypes() {
-        // TODO: Retrieve all of them.
-        // TODO: Retrieve all of import classes based on annotation in factory
         List<ImportConfig> configs = importManager.getAllTypeConfigs();
 
         //TODO: Add actual page size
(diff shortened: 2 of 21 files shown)


pom.xml
```diff
@@ -350,6 +350,11 @@
             <version>4.3</version>
             <scope>compile</scope>
         </dependency>
+        <dependency>
+            <groupId>org.reflections</groupId>
```
nickpalladino, line 354: Nice license on this library haha

src/main/java/org/breedinginsight/api/v1/controller/ImportController.java
```diff
@@ -0,0 +1,67 @@
...
+import javax.inject.Inject;
+import java.util.ArrayList;
+import java.util.List;
+
+@Slf4j
+@Controller("/${micronaut.bi.api.version}")
+public class ImportController {
+
+    @Inject
+    BrAPIImportConfigManager importManager;
```
nickpalladino, line 48: Maybe make this private and do constructor injection

src/main/java/org/breedinginsight/model/brapi_import/BrAPIImportConfigManager.java
```diff
@@ -0,0 +1,171 @@
...
+
+@Context
+@Singleton
+public class BrAPIImportConfigManager {
+
+    private Set<Class<?>> brAPIImports;
+
+    BrAPIImportConfigManager() {
+        // Get all imports
+        Reflections reflections = new Reflections("org.breedinginsight");
```
nickpalladino, line 43: Did you look at 3.16 Bean Annotation Metadata in the micronaut guide?

nickpalladino (commented):
What are you thinking for api docs? Skipping for now? Same question with tests. Nevermind, just read the PR text.

pom.xml
ctucker3: Hahah yeah Tim and I were looking at that too!

src/main/java/org/breedinginsight/model/brapi_import/BrAPIImportConfigManager.java
ctucker3: I did, but I can't remember if I tried it or not. I'll take another look.

PR #95, opened 2021-07-08
[BI-998] - Add tags to traits

ctucker3:
Some of the tests fail right now because tags were added and the brapi-server needs to be patched. This can still be review before then, I won't merge until the tests pass. 

Uses the brapi `variable.additionalInfo` field to store the tags. 

Additions:
  - POST traits with tags
  - PUT traits with tags
  - GET all trait tags within a program
  - Add tags to trait upload

Notes:
  - On the front end, existing tags assigned to the trait are based on lower case matching, and removing all spaces before matching. 
  
See other PRs for this feature:
brapi: https://github.com/Breeding-Insight/brapi/pull/40
sgn: https://github.com/Breeding-Insight/sgn/pull/30
bi-web: https://github.com/Breeding-Insight/bi-web/pull/99

LEFT TO DO: 
- Finish up tests in brapi library once Pete fixes the test server


src/main/java/org/breedinginsight/daos/TraitDAO.java
```diff
@@ -55,13 +56,17 @@
     @Property(name = "brapi.server.reference-source")
     private String referenceSource;
     private ObservationDAO observationDao;
+    private Gson gson;
+
+    private String TAGS_KEY = "tags";
```
nickpalladino, line 61: could be final static

src/main/java/org/breedinginsight/services/parsers/trait/TraitFileColumns.java
```diff
@@ -36,7 +36,8 @@
     SCALE_DECIMAL_PLACES("Scale decimal places"),
     SCALE_LOWER_LIMIT("Scale lower limit"),
     SCALE_UPPER_LIMIT("Scale upper limit"),
-    SCALE_CATEGORIES("Scale categories");
+    SCALE_CATEGORIES("Scale categories"),
+    TAGS("Tags");
```
nickpalladino, line 40: How is tags different from trait lists?

src/main/java/org/breedinginsight/services/TraitService.java
```diff
@@ -59,6 +59,8 @@
     private DSLContext dsl;
     private TraitValidatorError traitValidatorError;
 
+    private String FAVORITES_TAG = "favorites";
```
nickpalladino, line 62: could be final static

src/main/java/org/breedinginsight/services/TraitService.java
```diff
@@ -446,4 +448,21 @@ public Trait setTraitActiveStatus(UUID programId, UUID traitId, Boolean active,
...
+
+        // Parse out tags into unique list
+        Set<String> tags = new HashSet<>();
+        for (Trait trait: traits) {
+            if (trait.getTags() != null) {
+                tags.addAll(trait.getTags());
+            }
+        }
+
+        tags.add(FAVORITES_TAG);
```
nickpalladino, line 465: What is the purpose of this favorites tag?

ctucker3: There should always be a favorites tag as an option. This sends the favorites tag back in the options of tags even if a trait isn't assigned that tag yet.

nickpalladino (approved):

dmeidlin (approved):

PR #96, opened 2021-07-22
[BI-763] Updates to brapi v2 proxy to support Field Book

ctucker3:
- Changed the proxy endpoints to accept `HttpRequest<byte[]>`. It works the same, but this allows us to pass images through. 
- Passed through `Content-Type` through the proxy
- Fixed a bug that didn't account for `/brapi/v2` on the end of custom endpoints. 
- Broke up the proxy code a little more so I could reuse it for the `executeRequest` method and the new `executeByteRequest` method.

### DIFF
--- a/src/main/java/org/breedinginsight/brapi/v2/BrAPIV2Controller.java
+++ b/src/main/java/org/breedinginsight/brapi/v2/BrAPIV2Controller.java
@@ -159,7 +159,7 @@ public class BrAPIV2Controller {
         try {
             programBrAPIEndpoints = programService.getBrapiEndpoints(programId);
         } catch (DoesNotExistException e) {
-            throw new NotFoundException("Program does not exist");
+            throw new HttpStatusException(HttpStatus.NOT_FOUND, "Program does not exist");
         }
 
         if(programBrAPIEndpoints.getCoreUrl().isEmpty()) {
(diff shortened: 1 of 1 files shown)


timparsons (approved):

src/main/java/org/breedinginsight/brapi/v2/BrAPIV2Controller.java
```diff
@@ -79,81 +80,119 @@ public BrAPIServerInfoResponse serverinfo() {
...
+    }
 
-        throw new HttpStatusException(HttpStatus.UNAUTHORIZED, "Unauthorized BrAPI Request");
+    private HttpUrl getUrl(UUID programId, String path, HttpRequest<?> request) throws NotFoundException {
+
+        ProgramBrAPIEndpoints programBrAPIEndpoints;
+        try {
+            programBrAPIEndpoints = programService.getBrapiEndpoints(programId);
+        } catch (DoesNotExistException e) {
+            throw new NotFoundException("Program does not exist");
```
nickpalladino, line 162: Should this be an HttpStatusException with HttpStatus.NOT_FOUND?

ctucker3: Yeah probably, I'll add that.

ctucker3: changed

nickpalladino: can you get rid of the throws NotFoundException on the method signature?

ctucker3: removed

nickpalladino (approved):

PR #99, opened 2021-08-12
[BI-982] - Ontology BrAPI Mapping for Trait

dmeidlin:
The traitName for BrAPI traits is set to entity + attribute.
The traitName for BI traits is renamed to observationVariableName and is set to the value under the "Ontology term name" column for trait import.
Method names are autoformed from description + class

With these changes in place trait creation and editing will no longer work from the UI. Traits can only be added by batch importing until the UI for traits is updated.

Trait File Import:
- added "Trait description", "Trait entity", and "Trait attribute" columns
- renamed "Trait name" to "Ontology term name"
- added validators for the additional columns

Trait model:
- traitDescription field added

BI db:
- trait_name column in the trait table renamed to observation_variable_name(no data migration necessary)
- src code and tests updated to use new field name for traits.


timparsons (approved):
Looks good! Approving for the current code, but depending on the decision of the comment left on https://github.com/Breeding-Insight/bi-web/pull/106

nickpalladino:
> With these changes in place trait creation and editing will no longer work from the UI. Traits can only be added by batch importing until the UI for traits is updated.

Should we hold off on merging this card after it is approved until that card that addresses the UI side is done so we don't have broken traits on develop?

src/main/java/org/breedinginsight/services/parsers/trait/TraitFileParser.java
```diff
@@ -125,7 +125,7 @@ public TraitFileParser(TraitFileValidatorError traitValidatorError){
             ExcelRecord record = records.get(i);
 
             ProgramObservationLevel level = ProgramObservationLevel.builder()
-                    .name(parseExcelValueAsString(record, TraitFileColumns.TRAIT_LEVEL))
+                    .name(parseExcelValueAsString(record, TraitFileColumns.TRAIT_ENTITY))
```
nickpalladino, line 128: Is entity really the same as observation level or is it just being used for the autocomplete features of the program observation levels?

timparsons:
> > With these changes in place trait creation and editing will no longer work from the UI. Traits can only be added by batch importing until the UI for traits is updated.
> 
> Should we hold off on merging this card after it is approved until that card that addresses the UI side is done so we don't have broken traits on develop?

Probably not a bad idea

src/main/java/org/breedinginsight/services/validators/TraitValidatorError.java
```diff
@@ -69,8 +69,23 @@ public ValidationError getMissingScaleDataTypeMsg() {
...
     @Override
-    public ValidationError getMissingTraitNameMsg() {
-        return new ValidationError("traitName", "Missing trait name", HttpStatus.BAD_REQUEST);
+    public ValidationError getMissingObsVarNameMsg() {
+        return new ValidationError("observationVariableName", "Missing observation variable name", HttpStatus.BAD_REQUEST);
+    }
+
+    @Override
+    public ValidationError getMissingTraitEntityMsg() {
+        return new ValidationError("entity", "Missing trait entity", HttpStatus.BAD_REQUEST);
```
nickpalladino, line 78: Relating to previous comment, if the ObservationVariable was changed to have a trait object property then these would be trait.entity, trait.attribute, and trait.traitDescription to be consistent with how the other messages are formatted with the first part being the brapi object the properties are under. But that could be something done in another card.

src/main/java/org/breedinginsight/model/Trait.java
```diff
@@ -59,9 +59,9 @@
```
nickpalladino, line 59: If this class is really ObservationVariable, it may make sense to move the brapi trait object properties to a Trait object property like is being done for method and scale. Could be done in another card.

src/main/java/org/breedinginsight/services/parsers/trait/TraitFileParser.java
dmeidlin: just for the autocomplete features.
The card indirectly states that the two are not the same by saying that obs level should not be associated with traits.

src/main/java/org/breedinginsight/services/validators/TraitValidatorError.java
dmeidlin: agreed. There needs to be a major refactoring done(separate card) on the backend to go from using Trait as the highest level object (as it has been) to using observation variable as the highest level object(as you describe) to be consistent with the brapi model.
The scope of this work is functinality: making the changes necessary to get the backend to support the MVP changes made on the frontend. In the spirit of breaking the work up into manageable pieces done in a single sprint, I think the remaining necessary refactoring should be a separate card.

src/main/java/org/breedinginsight/brapi/v1/services/BrapiObservationVariableService.java
```diff
@@ -148,12 +148,12 @@ private ObservationVariable mapBiTraitToBrapiV1ObservationVariable(org.breedingi
                 //.status(trait.getStatus())
                 .synonyms(trait.getSynonyms())
                 .traitDbId(trait.getId().toString())
-                .traitName(trait.getTraitName())
+                .traitName(trait.getObservationVariableName())
```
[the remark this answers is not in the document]
dmeidlin: I believe it's correct as is.

In the Ontology MVP changes entity + attribute is mapped to the name of the BrAPI trait object, and the ontology term name (or just Name) is mapped to the BrAPI observation variable object, so 
observationVariable.trait.traitName = entity + attribute
observationVariable.observationVariableName = Name (which is stored in the Trait model as observationVariableName).

src/main/java/org/breedinginsight/brapi/v1/services/BrapiObservationVariableService.java
```diff
@@ -100,7 +100,7 @@ private ObservationVariable mapBiTraitToBrapiV1ObservationVariable(org.breedingi
         org.breedinginsight.model.Scale biScale = trait.getScale();
 
         Method brapiMethod = Method.builder()
-                .methodName(String.format("%s %s", trait.getTraitName(), biMethod.getMethodClass()))
+                .methodName(String.format("%s %s", trait.getObservationVariableName(), biMethod.getMethodClass()))
```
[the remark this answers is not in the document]
dmeidlin: Good catch! Yes, OV name wasn't intentional and should be description: fixed and pushed up.

src/main/java/org/breedinginsight/model/Trait.java
```diff
@@ -79,7 +79,6 @@ public Trait(TraitEntity traitEntity) {
         this.setId(traitEntity.getId());
         this.setMethodId(traitEntity.getMethodId());
         this.setScaleId(traitEntity.getScaleId());
-        this.setTraitName(traitEntity.getTraitName());
```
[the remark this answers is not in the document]
dmeidlin: agreed. done.

src/main/java/org/breedinginsight/model/Trait.java
dmeidlin: agreed. part of what I imagined would be included in the refactoring mentioned above.

src/main/java/org/breedinginsight/brapi/v1/services/BrapiObservationVariableService.java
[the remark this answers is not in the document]
dmeidlin: I looked at it again, you're right. line 151 is setting observationVariable.trait.traitName. will make the change.
Also will change line 146 should to `.description(trait.getTraitDescription).`

src/main/java/org/breedinginsight/services/validators/TraitValidatorError.java
nickpalladino: sounds good, could you create the card for that or at least a placeholder for it

src/main/java/org/breedinginsight/model/Trait.java
nickpalladino: sounds good

src/main/java/org/breedinginsight/services/parsers/trait/TraitFileColumns.java
```diff
@@ -22,10 +22,12 @@
 
 public enum TraitFileColumns {
 
-    TRAIT_NAME("Trait name"),
+    ONTOLOGY_TERM_NAME("Ontology term name"),
```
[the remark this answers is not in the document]
nickpalladino: Talked with @shawnyarnes and this will be addressed in BI-1059

src/main/java/org/breedinginsight/model/Trait.java
```diff
@@ -110,6 +109,7 @@ public static Trait parseSqlRecord(Record record) {
     public void setBrAPIProperties(BrAPIObservationVariable brApiVariable) {
         if (brApiVariable.getTrait() != null){
             this.setTraitClass(brApiVariable.getTrait().getTraitClass());
```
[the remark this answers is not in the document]
dmeidlin: There is a separate card for this: [BI-996](https://breedinginsight.atlassian.net/browse/BI-996).

src/main/java/org/breedinginsight/model/Trait.java
```diff
@@ -110,6 +109,7 @@ public static Trait parseSqlRecord(Record record) {
...
+            this.setObservationVariableName(brApiVariable.getObservationVariableName());
```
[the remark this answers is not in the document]
dmeidlin: Thanks, agreed.

nickpalladino (approved):

PR #100, opened 2021-08-13
Importer add studyType to Study

nickpalladino:
Breedbase `POST /studies` requires the studyType field to be set so adding it as a required field to the importer for the Study object.

### DIFF
--- a/src/main/java/org/breedinginsight/brapps/importer/model/base/Study.java
+++ b/src/main/java/org/breedinginsight/brapps/importer/model/base/Study.java
@@ -43,6 +43,17 @@ public class Study implements BrAPIObject {
 
     @ImportFieldType(type= ImportFieldTypeEnum.TEXT)
     @ImportMappingRequired
+    @ImportFieldMetadata(id="studyType", name="Study Type",
+            description = "Type of study, can be one of the predefined breedbase types or custom: " +
+                    "misc_trial, Preliminary Yield Trial, Uniform Yield Trial, heterosis_trial, " +
+                    "storage_trial, genetic_gain_trial, health_status_trial Seedling Nursery, " +
+                    "Variety Release Trial, Clonal Evaluation, Advanced Yield Trial, crossing_trial, " +
+                    "grafting_trial, pollinating_trial, phenotyping_trial, genotyping_trial, " +
+                    "Seed Multiplication, Screen House, crossing_block_trial, Specialty Trial, etc.")
+    private String studyType;
+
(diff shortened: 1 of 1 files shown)


src/main/java/org/breedinginsight/brapps/importer/model/base/Study.java
```diff
@@ -41,6 +41,17 @@
...
+    @ImportFieldType(type= ImportFieldTypeEnum.TEXT)
+    @ImportMappingRequired
+    @ImportFieldMetadata(id="studyType", name="Study Type",
+            description = "Type of study, can be one of the predefined breedbase types or custom: " +
+                    "misc_trial, Preliminary Yield Trial, Uniform Yield Trial, heterosis_trial, " +
+                    "storage_trial, genetic_gain_trial, health_status_trial Seedling Nursery, " +
+                    "Variety Release Trial, Clonal Evaluation, Advanced Yield Trial, crossing_trial, " +
+                    "grafting_trial, pollinating_trial, phenotyping_trial, genotyping_trial, " +
+                    "Seed Multiplication, Screen House, crossing_block_trial, Specialty Trial, etc.")
+    private String studyType;
```
ctucker3, line 53: I would remove `misc_trial` from the list. That's an internal breedbase thing.

ctucker3 (changes requested):

src/main/java/org/breedinginsight/brapps/importer/model/base/Study.java
nickpalladino: ya, I wasn't sure about that since it's showing up in the trial design as an option in the breedbase GUI. I grabbed all the terms from the db and looks like there's a few extra compared to the GUI.
image

HMS17 (approved):

src/main/java/org/breedinginsight/brapps/importer/model/base/Study.java
nickpalladino: Let me know if you still think I should remove it.

ctucker3: Oh I see. Yeah I had to put in a trial type of misc_trial whose cvterm value is recorded as the type the user passes in. In brapi it returns whatever the user enters and doesn't show misc_trial. So I guess we could show it, I just wonder if it is useful to anyone. I guess these help texts will all need a pass through by Shawn to clean them up anyway, so you can probably just leave it for now.

ctucker3 (approved):

PR #101, opened 2021-08-17
[BI-975] - Prevent creating programs with same name

HMS17:
Added check and error message to prevent creating programs with the same name.

This error message takes priority over the "BrAPI URL specified is not supported" error.

Added test to check for error, and modified existing tests to comply with unique program name

Connected to PR [bi-web/BI-975](https://github.com/Breeding-Insight/bi-web/pull/109)


src/main/java/org/breedinginsight/api/v1/controller/ProgramController.java
```diff
@@ -130,6 +130,9 @@ public ProgramController(ProgramService programService, ProgramUserService progr
             Program program = programService.create(programRequest, actingUser);
             Response<Program> response = new Response(program);
             return HttpResponse.ok(response);
+        } catch (AlreadyExistsException e){
```
nickpalladino, line 133: In some of the newer api code I've switched to using HttpStatusException then I don't have to catch anything in the controller and the HttpStatusExceptionHandler will handle it automatically. So for this you could throw a HttpStatusException(HttpStatus.CONFLICT, "Program name already in use") from the program service. Don't have to change this but just wanted to let you know for the future.

src/test/java/org/breedinginsight/api/v1/controller/ProgramControllerIntegrationTest.java
```diff
@@ -483,6 +484,30 @@ public void postProgramsInvalidSpecies() {
         assertEquals(HttpStatus.UNPROCESSABLE_ENTITY, e.getStatus());
     }
 
+    @Test
```
nickpalladino, line 487: It looks like validProgram is populated in a test with `@Order(2)` so it seems like this should be `@Order(3)`.

nickpalladino (commented):
We may want to clarify in the card acceptance criteria for testing that this will prevent duplicate program names in our db. In the event that there is an existing name in the brapi service but not our db, it will try to create it and in the case of breedbase we'll get an api exception back so it will fail but have a general failed to create program message in the UI rather than duplicate program.

src/main/java/org/breedinginsight/api/v1/controller/ProgramController.java
HMS17: Good to know. I was following the existing pattern given it already catches UnprocessableEntityException in the controller, but going forward I can use HttpStatusExceptionHandler.

src/test/java/org/breedinginsight/api/v1/controller/ProgramControllerIntegrationTest.java
HMS17: Good point, updated.

nickpalladino (approved):

src/main/java/org/breedinginsight/services/ProgramService.java
```diff
@@ -223,4 +229,14 @@ public ProgramBrAPIEndpoints getBrapiEndpoints(UUID programId) throws DoesNotExi
         return dao.getProgramBrAPIEndpoints(programId);
     }
 
+    private boolean programNameInUse(String name) {
+        List<ProgramEntity> existingPrograms = dao.fetchByName(name);
```
ctucker3, line 233: Is this check case-insensitive? Do we want a case insensitive match?

ctucker3 (changes requested):

src/main/java/org/breedinginsight/services/ProgramService.java
HMS17: > Is this check case-insensitive? Do we want a case insensitive match?

I think it's case-sensitive. I'll ask in the channel about a case insensitive match.

ctucker3 (approved):

src/main/java/org/breedinginsight/services/ProgramService.java
HMS17: > > Is this check case-insensitive? Do we want a case insensitive match?
> 
> I think it's case-sensitive. I'll ask in the channel about a case insensitive match.

We're going with case insensitive so I modified it accordingly.

PR #105, opened 2021-09-29
[BI-1059] Name vs Full Name

HMS17:
Backend changes for Ontology Name and Full Name:

- Add mapping for Name
- Add mapping for Full Name
- Remove required field validations for Method Description
- Add character limit validation for Name, Entity, Attribute, and Method Description
- Add validation that there are at least two filled fields if ordinal scale
- Add validation that there are at least two filled fields if nominal scale

Linked to BI-1127 (which handles front end validation of ordinal and nominal scale)

### DIFF
--- a/src/main/java/org/breedinginsight/services/validators/TraitFileValidatorError.java
+++ b/src/main/java/org/breedinginsight/services/validators/TraitFileValidatorError.java
@@ -130,6 +130,26 @@ public class TraitFileValidatorError implements TraitValidatorErrorInterface {
     }
 
     @Override
+    public ValidationError getCharLimitObsVarNameMsg() {
+        return new ValidationError("observationVariableName", "Observation variable name exceeds 12 character limit", HttpStatus.UNPROCESSABLE_ENTITY);
+    }
+
+    @Override
+    public ValidationError getCharLimitTraitEntityMsg() {
+        return new ValidationError("entity", "Trait entity exceeds 30 character limit", HttpStatus.UNPROCESSABLE_ENTITY);
+    }
+
+    @Override
+    public ValidationError getCharLimitTraitAttributeMsg() {
+        return new ValidationError("attribute", "Trait attribute exceeds 30 character limit", HttpStatus.UNPROCESSABLE_ENTITY);
+    }
+
+    @Override
+    public ValidationError getCharLimitMethodDescriptionMsg() {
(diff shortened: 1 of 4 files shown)


src/main/java/org/breedinginsight/daos/TraitDAO.java
```diff
@@ -419,7 +422,7 @@ public Trait updateTraitBrAPI(Trait trait, Program program) {
                 existingVariable.setStatus("archived");
             }
             existingVariable.putAdditionalInfoItem(TAGS_KEY, trait.getTags());
-
+            existingVariable.putAdditionalInfoItem(FULLNAME_KEY, trait.getFullName());
```
nickpalladino, line 425: Should this also not write the additionalInfo item if fullName is null like the create call?

src/main/java/org/breedinginsight/services/validators/TraitFileValidatorError.java
```diff
@@ -129,6 +124,36 @@ public ValidationError getMaxLessThenMinError() {
                 HttpStatus.UNPROCESSABLE_ENTITY);
     }
 
+    @Override
```
nickpalladino, line 127: Didn't see these 2 cases in card, maybe should be added to acceptance criteria

src/main/java/org/breedinginsight/services/validators/TraitValidatorService.java
```diff
@@ -166,6 +171,38 @@ public ValidationErrors checkTraitDataConsistency(List<Trait> traits, TraitValid
         return errors;
     }
 
+    public ValidationErrors checkTraitFieldsLength(List<Trait> traits, TraitValidatorErrorInterface traitValidatorErrors) {
```
nickpalladino, line 174: Doesn't look like this is actually called as part of the trait error checking in checkAllTraitValidations.

src/main/java/org/breedinginsight/daos/TraitDAO.java
HMS17: Added the precaution just in case and checked that it didn't prevent clearing the fullname on an edit.

src/main/java/org/breedinginsight/services/validators/TraitValidatorService.java
HMS17: Good catch, added in most recent commit.

nickpalladino (approved):

PR #113, opened 2021-11-10
[features/BI-1139] added Flyway for breading_methods table

davedrp:
Adds and populates breading_methods table


HMS17 (approved):

src/main/resources/db/migration/V0.5.24__create-breeding_method-table.sql
```diff
@@ -0,0 +1,91 @@
+/*
```
HMS17, line 1: Looks like you need to change the file name to V0.5**.25**__create-breeding_method-table.sql in order to get it to pass the checks.

HMS17 (changes requested):

src/main/resources/db/migration/V0.5.24__create-breeding_method-table.sql
davedrp: thx

HMS17 (approved):

dmeidlin (approved):

PR #115, opened 2021-11-17
[BI-1223] Add null check for tags field

ctucker3:
The real fix to this was to update the reltest brapi server database. This fix is to make it so that there is a bug that returns something like:

```
additionalInfo: {
  tags: null
}
```

from the brapi-server again, it won't prevent the user from seeing their list of traits.

### DIFF
--- a/src/main/java/org/breedinginsight/daos/TraitDAO.java
+++ b/src/main/java/org/breedinginsight/daos/TraitDAO.java
@@ -554,9 +554,13 @@ public class TraitDAO extends TraitDao {
 
     private void saturateTrait(Trait trait, BrAPIObservationVariable brApiVariable) {
 
-        if (brApiVariable.getAdditionalInfo() != null && brApiVariable.getAdditionalInfo().has(TAGS_KEY)) {
-            List<String> tags = gson.fromJson(brApiVariable.getAdditionalInfo().getAsJsonArray(TAGS_KEY), List.class);
-            trait.setBrAPIProperties(brApiVariable, tags);
+        if (brApiVariable.getAdditionalInfo() != null &&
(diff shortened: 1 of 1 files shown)


HMS17 (approved):

src/main/java/org/breedinginsight/daos/TraitDAO.java
```diff
@@ -554,9 +554,13 @@ private Trait parseTraitRecord(Record record, BiUserTable createdByUser, BiUserT
 
     private void saturateTrait(Trait trait, BrAPIObservationVariable brApiVariable) {
 
-        if (brApiVariable.getAdditionalInfo() != null && brApiVariable.getAdditionalInfo().has(TAGS_KEY)) {
-            List<String> tags = gson.fromJson(brApiVariable.getAdditionalInfo().getAsJsonArray(TAGS_KEY), List.class);
-            trait.setBrAPIProperties(brApiVariable, tags);
+        if (brApiVariable.getAdditionalInfo() != null &&
+                brApiVariable.getAdditionalInfo().has(TAGS_KEY)) {
+            // Prevents BrAPI server from failing the tags load
+            if (!brApiVariable.getAdditionalInfo().get(TAGS_KEY).isJsonNull()) {
```
nickpalladino, line 560: Does there need to be a check in the create/update calls so that null is not be written for the tags additional info value?

ctucker3: Those work fine right now for saving a new trait with no tag and updating a trait to remove all of the tags, so I think we are good with leaving those alone.

nickpalladino (approved):

PR #116, opened 2021-11-19
[BI-1176] Create Germplasm Template Mapping

HMS17:
- Migration to add default system level germplasm template mapping
- Update validation tests to account for existing system mapping

Linked to PR [bi-web/BI-1176](https://github.com/Breeding-Insight/bi-web/pull/150)

### DIFF
--- a/src/test/java/org/breedinginsight/api/v1/controller/ImportControllerIntegrationTest.java
+++ b/src/test/java/org/breedinginsight/api/v1/controller/ImportControllerIntegrationTest.java
@@ -156,8 +156,16 @@ public class ImportControllerIntegrationTest extends BrAPITest {
         JsonObject result = JsonParser.parseString(response.body()).getAsJsonObject().getAsJsonObject("result");
         JsonArray data = result.getAsJsonArray("data");
 
+        boolean importNameFound = false;
         assertEquals(2, data.size(), "Wrong number of results returned.");
-        assertEquals("GermplasmTest", data.get(1).getAsJsonObject().get("name").getAsString(), "Wrong import name returned");
+        for (JsonElement jsonMapping: data) {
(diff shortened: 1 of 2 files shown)


src/test/java/org/breedinginsight/api/v1/controller/ImportControllerIntegrationTest.java
```diff
@@ -156,8 +157,16 @@ public void getSystemMappings() {
         JsonObject result = JsonParser.parseString(response.body()).getAsJsonObject().getAsJsonObject("result");
         JsonArray data = result.getAsJsonArray("data");
 
-        assertEquals(1, data.size(), "Wrong number of results returned.");
-        assertEquals("GermplasmTest", data.get(0).getAsJsonObject().get("name").getAsString(), "Wrong import name returned");
+        boolean importNameFound = false;
```
nickpalladino, line 160: We may want to update this test so it doesn't fail each time a new system mapping is added. It could just look for the test mapping potentially and there could be other tests for each specific system mapping.

HMS17: That is a good point. So for now just remove the line looking for the size of the system mapping list?

nickpalladino: Ya you could do that. You could also add another test if you wanted using the name query parameter to verify that the germplasm system template exists.

HMS17: Alright, implemented!

nickpalladino (approved):

dmeidlin (approved):

PR #120, opened 2021-12-01
[BI-1246] Program Key Uniqueness Error Fix

HMS17:
- Fixed error message when user tries to enter a program key that is already in use

Linked to PR [taf/BI-1246](https://github.com/Breeding-Insight/taf/pull/37)


dmeidlin (approved):

src/main/java/org/breedinginsight/services/ProgramService.java
```diff
@@ -57,7 +57,7 @@
     private BrAPIClientProvider brAPIClientProvider;
 
     private static final String PROGRAM_NAME_IN_USE = "PROGRAM_NAME_IN_USE";
-    private static final String PROGRAM_KEY_IN_USE = "PROGRAM_NAME_IN_USE";
+    private static final String PROGRAM_KEY_IN_USE = "PROGRAM_KEY_IN_USE";
```
nickpalladino, line 60: Is there a reason these strings are formatted this way rather than just normal human readable text? Guessing you were just going off program name, but I think everything else just uses normal english messages.

HMS17: Mirroring program name, yes. I think it might also be because that value is used to determine what type of message to return in ProgramService.ts rather than directly giving the message.

```
} else if (error.response && error.response.status === 409 && error.response.statusText === ProgramService.PROGRAM_KEY_IN_USE) {
            error['errorMessage'] = this.duplicateProgramKey;
```
where PROGRAM_KEY_IN_USE would have been returned as the error.response.statusText

nickpalladino: ah, thanks for the clarification

nickpalladino (approved):

PR #131, opened 2022-01-06
BI-1143 - Method Class column in Ontology Import file allows any value on import

nickpalladino:
# Description
**Story:** [BI-1143](https://breedinginsight.atlassian.net/browse/BI-1143?atlOrigin=eyJpIjoiNmRiOTE0MGU5M2I5NDk1MmIxYjU4NTA2YjgzZWE1ZGYiLCJwIjoiaiJ9)

Added check in file parser to only allow BI supported method classes, not any string.

# Dependencies
Just develop on bi-api & bi-web.

# Testing
Updated `putTraitUploadBadTypes` integration test/file to demonstrate check works.

# Note
Went with an enum but didn't change in Method model from String because stuff may need to be migrated and not sure we want to limit. 

# Checklist:

- [x] I have made corresponding changes to documentation
- [ ] I have run TAF.


ctucker3 (approved):
Tested with a bad method class and a good method class, both behaved as expected.

HMS17 (approved):

PR #150, opened 2022-02-09
[BI-1281] refactor obsVarName error messages to use Name

dmeidlin:
# Description
**Story:** [BI-1281](https://breedinginsight.atlassian.net/jira/software/c/projects/BI/boards/1?modal=detail&selectedIssue=BI-1281)

trait import validation error messages for observation variable names were changed to refer instead to "name"

# Dependencies
bi-web `develop`

# Testing
try importing a trait with a name that exceeds the 12 char limit.

# Checklist:

- [ ] I have run TAF: _\<link to TAF run>_

### DIFF
--- a/src/main/java/org/breedinginsight/services/validators/TraitValidatorError.java
+++ b/src/main/java/org/breedinginsight/services/validators/TraitValidatorError.java
@@ -67,7 +67,7 @@ public class TraitValidatorError implements TraitValidatorErrorInterface {
 
     @Override
     public ValidationError getMissingObsVarNameMsg() {
-        return new ValidationError("observationVariableName", "Missing observation variable name", HttpStatus.BAD_REQUEST);
+        return new ValidationError("name", "Missing Name", HttpStatus.BAD_REQUEST);
     }
 
     @Override
@@ -140,7 +140,7 @@ public class TraitValidatorError implements TraitValidatorErrorInterface {
 
     @Override
     public ValidationError getCharLimitObsVarNameMsg() {
-        return new ValidationError("observationVariableName", "Observation variable name exceeds 12 character limit", HttpStatus.UNPROCESSABLE_ENTITY);
+        return new ValidationError("name", "Name exceeds 12 character limit", HttpStatus.UNPROCESSABLE_ENTITY);
     }
 
     @Override
--- a/src/main/java/org/breedinginsight/services/validators/TraitValidatorError.java
+++ b/src/main/java/org/breedinginsight/services/validators/TraitValidatorError.java
@@ -37,62 +37,62 @@ public class TraitValidatorError implements TraitValidatorErrorInterface {
 
     @Override
     public ValidationError getTraitIdDoesNotExistMsg() {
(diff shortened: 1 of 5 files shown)


HMS17 (changes requested):
Tested with a file with name and description exceeding character limits, got:
"Name: Name exceeds 12 character limit in row 2
method.description: Method description exceeds 30 character limit in row 2"

Based on the card, the second line should be more akin to "Method Description: Method description exceeds 30 character limit in row x", correct?

Also tested a scenario with insufficient scale categories and got "scale.categories: Ordinal scales must have at least two categories. in row 4" which matches the expected result but I'm wondering if that is a typo given that seems more the variable name than the column name.

nickpalladino (changes requested):
I think this should target the v0.5 release branch right?

HMS17 (approved):

src/main/java/org/breedinginsight/services/validators/TraitValidatorError.java
```diff
@@ -67,7 +67,7 @@ public ValidationError getMissingScaleDataTypeMsg() {
 
     @Override
     public ValidationError getMissingObsVarNameMsg() {
-        return new ValidationError("observationVariableName", "Missing observation variable name", HttpStatus.BAD_REQUEST);
+        return new ValidationError("name", "Missing Name", HttpStatus.BAD_REQUEST);
```
nickpalladino, line 70: Should this be capitalized?

nickpalladino: Same question for entity and attribute in here and TraitValidationError.

src/main/java/org/breedinginsight/services/validators/TraitValidatorError.java
```diff
@@ -155,7 +155,7 @@ public ValidationError getCharLimitTraitAttributeMsg() {
 
     @Override
     public ValidationError getCharLimitMethodDescriptionMsg() {
-        return new ValidationError("method.description", "Method description exceeds 30 character limit", HttpStatus.UNPROCESSABLE_ENTITY);
+        return new ValidationError("Method Description", "Method description exceeds 30 character limit", HttpStatus.UNPROCESSABLE_ENTITY);
     }
 
     @Override
```
nickpalladino, line 161: Should the terminology here and in TraitValidationError be changed to name rather than trait name? Should it also be capitalized?

src/main/java/org/breedinginsight/services/validators/TraitValidatorError.java
nickpalladino: Should trait traitDescription, programObservationLevel.name, and method.formula be updated?

src/main/java/org/breedinginsight/services/validators/TraitValidatorError.java
dmeidlin: Trait name is now a concatenation of entity + attribute; whereas, obs var. name is the Name. So my feeling is to keep trait name as is for now, but I updated the rest of the field names to have consistent formatting/capitalization.

nickpalladino (approved):

PR #243, opened 2023-03-03
[BI-1629] Add filtering to Ontology Table

dmeidlin:
# Description
**Story:** [BI-1629](https://breedinginsight.atlassian.net/jira/software/c/projects/BI/boards/1?modal=detail&selectedIssue=BI-1629)

`TraitsQuery#constructSearchRequest` was created and called in the GET /traits controller of `TraitsController`.
The query params and search request are then passed to the same `ResponseUtils` method used by POST /traits for returning filtered traits.

# Dependencies
none

# Testing
call `GET /traits?<field>=<value>` using a searchable field for traits and verify only filtered traits returned.

# Checklist:

[BI-1629]: https://breedinginsight.atlassian.net/browse/BI-1629?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/org/breedinginsight/api/model/v1/request/query/TraitsQuery.java
+++ b/src/main/java/org/breedinginsight/api/model/v1/request/query/TraitsQuery.java
@@ -19,9 +19,117 @@ package org.breedinginsight.api.model.v1.request.query;
 
 import io.micronaut.core.annotation.Introspected;
 import lombok.Getter;
+import org.apache.commons.lang3.StringUtils;
+import java.util.ArrayList;
+import java.util.List;
 
 @Getter
 @Introspected
 public class TraitsQuery extends QueryParams {
+    private String name;
+    private String traitDescription;
+    private String entityAttribute;
+    private String mainAbbreviation;
+    private String synonyms;
+    private String level;
+    private String status;
+    private String methodDescription;
+    private String methodClass;
+    private String methodFormula;
+    private String scaleName;
+    private String scaleClass;
+    private String scaleDecimalPlaces;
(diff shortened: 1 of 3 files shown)


src/main/java/org/breedinginsight/api/model/v1/request/query/TraitsQuery.java
```diff
@@ -19,9 +19,117 @@
...
+        }
+        if (!StringUtils.isBlank(getStatus())) {
+            filters.add(constructFilterRequest("status", getStatus()));
+        }
+        if (!StringUtils.isBlank(getMethodDescription())) {
+            filters.add(constructFilterRequest("methodDescription", getMethodDescription()));
+        }
+        if (!StringUtils.isBlank(getMethodClass())) {
+            filters.add(constructFilterRequest("methodClass", getMethodClass()));
+        }
```
timparsons, line 81: I think these two fields need to be combined together for filtering purposes because the UI combines them together for display purposes.  When I tried filtering by that column, the filter only applies to the method description.  

Example UI filtering: on the UI the "Method" column shows "test Measurement".  If I type in "test" as my filter, then the ontology term shows correctly, however, if I type "Measurement", the term does not show in the table.

timparsons (changes requested):
Code looks good, but had one issue when I was testing from the UI

davedrp (approved):

timparsons (approved):

PR #244, opened 2023-03-13
BI-1741 - Genotype Import

timparsons:
# Description
**Story:** https://breedinginsight.atlassian.net/browse/BI-1741

Updating how calls are fetched to prevent duplication of calls when there is more than one callset per germplasm

# Dependencies
bi-web bug/BI-1741
brapi bug/BI-1741
Mgdb2BrapiV2Impl bug/BI-1741

# Testing
Test using instructions from [BI-1650](https://breedinginsight.atlassian.net/browse/BI-1650)

# Checklist:

- [x] I have made corresponding changes to documentation

[BI-1650]: https://breedinginsight.atlassian.net/browse/BI-1650?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/org/breedinginsight/utilities/BrAPIDAOUtil.java
+++ b/src/main/java/org/breedinginsight/utilities/BrAPIDAOUtil.java
@@ -167,7 +167,7 @@ public class BrAPIDAOUtil {
             //NOTE: Because of the way Breedbase implements BrAPI searches, the page size is initially set to an
             //arbitrary, large value to ensure that in the event that a 202 response is returned, the searchDbId
             //stored will refer to all records of the BrAPI variable.
-            searchBody.pageSize(1000);
+            searchBody.pageSize(pageSize);
             ApiResponse<Pair<Optional<T>, Optional<BrAPIAcceptedSearchResponse>>> response = searchMethod.apply(searchBody);
             if (response.getBody().getLeft().isPresent()) {
                 BrAPIResponse listResponse = (BrAPIResponse) response.getBody().getLeft().get();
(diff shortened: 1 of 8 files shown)


dmeidlin (approved):

src/main/java/org/breedinginsight/utilities/BrAPIDAOUtil.java
```diff
@@ -167,7 +167,7 @@ public <T, U extends BrAPISearchRequestParametersTokenPaging, V> List<V> searchW
             //NOTE: Because of the way Breedbase implements BrAPI searches, the page size is initially set to an
             //arbitrary, large value to ensure that in the event that a 202 response is returned, the searchDbId
             //stored will refer to all records of the BrAPI variable.
-            searchBody.pageSize(10000000);
+            searchBody.pageSize(1000);
```
davedrp, line 170: I can see why 10000000 is excessively large.  But is 1000 large enough?  Too small of a pageSize could cause a problem, but is there a down side to a pageSize that it too large?

davedrp (changes requested):

src/main/java/org/breedinginsight/utilities/BrAPIDAOUtil.java
timparsons: @davedrp Fair question.  It could be, but I think part of that would be doing some trial and error.  For calls to BreedBase, we had to set this to an excessively large number due to the way in which the `POST /search` endpoints are implemented.  For Gigwa, there's a bug where they limit the page size, but the pagination metadata in the response is still utilizing the page size sent on the request to calculate the total number of pages.  Because of this, I changed the page size to `1000` for the `searchWithToken` method (that method is only utilized by the code interacting with Gigwa at the moment).

I updated this to use the `brapi.page-size` configuration that's in `application.yml` so

davedrp (approved):

PR #252, opened 2023-05-09
[BI-1689] - Incorrect Jobs Count

mlm483:
# Description
**Story:** https://breedinginsight.atlassian.net/browse/BI-1689

There were 2 bugs causing the "Complete count" (a.k.a. finished count) to be less than the "Total count" on the jobs page:

1. The first bug was at the end of the `BrAPIDAOUtil.post` method where it was updating the finished count to be `listResult.size()`. The problem was that `listResult` contains all the BrAPI objects created during that call to `post`, but the finished count needs to account for BrAPI objects created during previous calls to `post`. I fixed this by using the local variable `finished` which is initially set to the current finished count (thus accounting for previous calls to `post`) and then incremented after each BrAPI call.
2. The object counts that are generated when a file is parsed, e.g. by `GermplasmProcessor.getStatisticsMap`, count related objects (e.g. pedigree for germplasm). These related objects are not POSTed separately to the BrAPI service, so there isn't a generic way to keep track of them, they would need to be tracked differently for each BrAPI object type. An expedient solution was to update the finished count to account for these related objects at the end of the `ProcessorManager.postBrapiData` method ([Line 118](https://github.com/Breeding-Insight/bi-api/blob/e31b705b0b6e54ad54ac881a144383614ccc6fe3/src/main/java/org/breedinginsight/brapps/importer/services/processors/ProcessorManager.java#L118)). There is already logic to make sure the expected number of primary BrAPI objects matches the number created in `BrAPIDAOUtil.post` ([Lines 322-324](https://github.com/Breeding-Insight/bi-api/blob/e31b705b0b6e54ad54ac881a144383614ccc6fe3/src/main/java/org/breedinginsight/utilities/BrAPIDAOUtil.java#L322)).

# Checklist:

- [x] I have run TAF: https://github.com/Breeding-Insight/taf/actions/runs/4982646479 (2nd run succeeded)

### DIFF
--- a/src/main/java/org/breedinginsight/brapps/importer/services/processors/ProcessorManager.java
+++ b/src/main/java/org/breedinginsight/brapps/importer/services/processors/ProcessorManager.java
@@ -114,6 +114,9 @@ public class ProcessorManager {
             processor.postBrapiData(mappedBrAPIImport, program, upload);
         }
 
+        // Update progress to reflect final finished and inProgress counts.
+        upload.updateProgress(Math.toIntExact(totalObjects), 0);
+
         log.debug("Completed upload to brapi service");
         statusService.finishUpload(upload, "Completed upload to brapi service");
     }
--- a/src/main/java/org/breedinginsight/brapps/importer/services/processors/ProcessorManager.java
+++ b/src/main/java/org/breedinginsight/brapps/importer/services/processors/ProcessorManager.java
@@ -114,11 +114,8 @@ public class ProcessorManager {
             processor.postBrapiData(mappedBrAPIImport, program, upload);
         }
 
-        // Update progress to reflect final finished and inProgress counts.
-        upload.updateProgress(Math.toIntExact(totalObjects), 0);
-
         log.debug("Completed upload to brapi service");
-        statusService.finishUpload(upload, "Completed upload to brapi service");
+        statusService.finishUpload(upload, totalObjects, "Completed upload to brapi service");
     }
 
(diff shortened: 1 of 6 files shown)


src/main/java/org/breedinginsight/brapps/importer/services/processors/ProcessorManager.java
```diff
@@ -114,6 +114,9 @@ private void postBrapiData(Program program, ImportUpload upload) throws Validato
             processor.postBrapiData(mappedBrAPIImport, program, upload);
         }
 
+        // Update progress to reflect final finished and inProgress counts.
+        upload.updateProgress(Math.toIntExact(totalObjects), 0);
```
nickpalladino, line 118: Could this be moved into the ImportStatusService::finishUpload method? Mainly just an organizational thing to keep all modifications to the ImportUpload object being done in there.

nickpalladino (changes requested):
Discussed newObject counts and maybe in the future could be cleaned up so that total only includes brapi objects being created or updated.

src/main/java/org/breedinginsight/brapps/importer/services/processors/ProcessorManager.java
mlm483: Done and tested, b09ddeaf.

nickpalladino (approved):

dmeidlin (approved):

PR #284, opened 2023-08-15
[BI-1845] - add listType parameter to GET /lists calls

mlm483:
# Description
**Story:** https://breedinginsight.atlassian.net/browse/BI-1845

_Please include a summary of the change that was made._

# Dependencies
_Please include any dependencies to other code branches, testing configurations, scripts to be run, etc._

# Testing
_Please include any details needed for reviewers to test this code_

# Checklist:

### DIFF
--- a/src/main/java/org/breedinginsight/brapi/v2/ListController.java
+++ b/src/main/java/org/breedinginsight/brapi/v2/ListController.java
@@ -79,6 +79,7 @@ public class ListController {
 
             // get germplasm lists by default
             BrAPIListTypes type = BrAPIListTypes.fromValue(queryParams.getListType());
+            // TODO: return 400 bad request if no listDbid or listType.
             if (type == null) {
                 type = BrAPIListTypes.GERMPLASM;
             }


src/main/java/org/breedinginsight/brapi/v2/ListController.java
```diff
@@ -80,7 +80,7 @@ public HttpResponse<Response<DataResponse<Object>>> getLists(
             // get germplasm lists by default
             BrAPIListTypes type = BrAPIListTypes.fromValue(queryParams.getListType());
             if (type == null) {
-                type = BrAPIListTypes.GERMPLASM;
```
timparsons, line 83: I think that this goes against the BrAPI spec.  I think we should probably return all lists if the `type` parameter is `null`.  Worth confirming with @BrapiCoordinatorSelby

timparsons (changes requested):

src/main/java/org/breedinginsight/brapi/v2/ListController.java
BrapiCoordinatorSelby: The BrAPI Spec does not specifically define what should happen if `listType` is null. 
I would expect it to return all lists of all types, in line with other GET query parameters. You might also make `listType` a required parameter for your implementation. 
Defaulting `listType` to germplasm is unusual, and might cause some confusion if it is not well documented, but it is not technically against the BrAPI spec.

mlm483: Thanks for the input @BrapiCoordinatorSelby, all lists of all types are now returned.

src/main/java/org/breedinginsight/brapi/v2/services/BrAPIListService.java
```diff
@@ -72,19 +72,20 @@ public List<BrAPIListSummary> getListSummariesByTypeAndXref(
...
-                case GERMPLASM:
-                default:
-                    String createdBy = germplasmDAO.getGermplasmByRawName(listItemNames, program.getId()).get(0)
-                            .getAdditionalInfo()
-                            .getAsJsonObject("createdBy")
-                            .get("userName")
-                            .getAsString();
-                    list.setListOwnerName(createdBy);
+            if (type != null) {
+                switch (type) {
```
mlm483, line 76: Prevents `NullPointerException` for switch on null object.

src/main/java/org/breedinginsight/brapi/v2/services/BrAPIListService.java
```diff
@@ -72,19 +72,20 @@ public List<BrAPIListSummary> getListSummariesByTypeAndXref(
...
+                        break;
+                    case GERMPLASM:
+                    default:
+                        String createdBy = germplasmDAO.getGermplasmByRawName(listItemNames, program.getId()).get(0)
+                                .getAdditionalInfo()
+                                .getAsJsonObject("createdBy")
+                                .get("userName")
+                                .getAsString();
+                        list.setListOwnerName(createdBy);
+                }
```
timparsons, line 87: I know this wasn't a change you made, but this code should probably be moved inside the `GERMPLASM` case

timparsons (approved):
One suggested change to existing code, but doesn't need to stop my approval of the main change.

src/main/java/org/breedinginsight/brapi/v2/services/BrAPIListService.java
mlm483: Done, d029f0d6.

A switch doesn't really make sense at the moment, but I assume it will get more use in the future.

davedrp (approved):

PR #288, opened 2023-08-25
BI-1881 (Remove Need to Provide Parents to processGermplasmForDisplay)

davedrp:
# Description
[BI-1881](https://breedinginsight.atlassian.net/browse/BI-1881) Remove Need to Provide Parents to processGermplasmForDisplay

- Any call to processGermplasmForDisplay() now will only need to pass in a list of the modified germplasm, not the entire list of germplasm for the program
- if a parent of a gerplasm is not referenced on the passed-in list of germplasm, then, and only then will the code fetch the entire list of germplasm for the program.

# Checklist:

- [X] I have tested that my code works with both the brapi-java-server and BreedBase
- [X] I have run TAF: [please include a link to TAF run](https://github.com/Breeding-Insight/taf/actions/runs/5978812089)

[BI-1881]: https://breedinginsight.atlassian.net/browse/BI-1881?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/org/breedinginsight/brapi/v2/dao/BrAPIGermplasmDAO.java
+++ b/src/main/java/org/breedinginsight/brapi/v2/dao/BrAPIGermplasmDAO.java
@@ -35,6 +35,7 @@ import org.breedinginsight.brapi.v2.constants.BrAPIAdditionalInfoFields;
 import org.breedinginsight.brapps.importer.daos.ImportDAO;
 import org.breedinginsight.brapps.importer.model.ImportUpload;
 import org.breedinginsight.brapps.importer.services.ExternalReferenceSource;
+import org.breedinginsight.dao.db.tables.pojos.ProgramEntity;
 import org.breedinginsight.daos.ProgramDAO;
 import org.breedinginsight.daos.cache.ProgramCache;
 import org.breedinginsight.daos.cache.ProgramCacheProvider;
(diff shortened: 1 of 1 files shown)


src/main/java/org/breedinginsight/brapi/v2/dao/BrAPIGermplasmDAO.java
```diff
@@ -241,6 +259,16 @@ private Map<String,BrAPIGermplasm> processGermplasmForDisplay(List<BrAPIGermplas
         return programGermplasmMap;
     }
 
+    private boolean expand_programGermplasmByFullName(ProgramEntity program, Map<String, BrAPIGermplasm> programGermplasmByFullName) throws ApiException {
```
mlm483, line 262: Unless the underscore in the name is part of a convention I don't know about, I would remove it.

src/main/java/org/breedinginsight/brapi/v2/dao/BrAPIGermplasmDAO.java
```diff
@@ -145,21 +146,23 @@ private Map<String, BrAPIGermplasm> fetchProgramGermplasm(UUID programId) throws
...
-    private Map<String,BrAPIGermplasm> processGermplasmForDisplay(List<BrAPIGermplasm> programGermplasm, String programKey) {
+    private Map<String,BrAPIGermplasm> processGermplasmForDisplay(List<BrAPIGermplasm> germplasmList, ProgramEntity program) throws ApiException {
+        String programKey = program.getKey();
         // Process the germplasm
         Map<String, BrAPIGermplasm> programGermplasmMap = new HashMap<>();
-        log.trace("processing germ for display: " + programGermplasm);
+        log.trace("processing germ for display: " + germplasmList);
         Map<String, BrAPIGermplasm> programGermplasmByFullName = new HashMap<>();
-        for (BrAPIGermplasm germplasm: programGermplasm) {
+        boolean isExpanded_programGermplasmByFullName = false;
```
mlm483, line 164: Unless the underscore in the name is part of a convention I don't know about, I would remove it.
`isExpandedProgramGermplasmByFullName` would be OK with me, something like `isMissingParents` would be OK too.

src/main/java/org/breedinginsight/brapi/v2/dao/BrAPIGermplasmDAO.java
```diff
@@ -241,6 +259,16 @@ private Map<String,BrAPIGermplasm> processGermplasmForDisplay(List<BrAPIGermplas
         return programGermplasmMap;
     }
 
+    private boolean expand_programGermplasmByFullName(ProgramEntity program, Map<String, BrAPIGermplasm> programGermplasmByFullName) throws ApiException {
+        List<BrAPIGermplasm> allProgramGermplasm = getRawGermplasm(program.getId());
```
mlm483, line 263: `getRawGermplasm` hits the cache, calling it from the method `processGermplasmForDisplay` which is used to populate the cache may be causing the issues I'm seeing locally and in the TAF run.

mlm483 (changes requested):
Looking at the failures in the TAF run and testing locally, I'm noticing that uploading Germplasm in a program ***that doesn't have any Germplasm yet*** fails. It will hang on after the user clicks "Upload" or "Confirm", I think depending on when the cache gets hit.

src/main/java/org/breedinginsight/brapi/v2/dao/BrAPIGermplasmDAO.java
nickpalladino: I thought the original intent of this card was to grab the accession numbers from the pedigree string [programKey-accessionNumber] and use that to create the newPedigreeString rather than requiring the parent germplasm objects @timparsons? Although I'm not sure how it would create the name and UUID pedigree strings in that case.

nickpalladino (changes requested):

mlm483:
Closing in favor of https://github.com/Breeding-Insight/bi-api/pull/290

PR #320, opened 2023-12-08
[BI-2028] Duplicated accession names getting assigned duplicated GIDs after Exp&Obs upload

dmeidlin:
# Description
**Story:** [BI-2028](https://breedinginsight.atlassian.net/browse/BI-2028)

During experiment import, existing germplasm dbid was being assigned to OU PIO by looking up the germplasm name stored in the OU PIO. When there are multiple germplasm with the same name, this was causing the same germplasm to be assigned to multiple OUs.

- Changed the OU PIO creation to include the germplasm GID as additional info
- mapped the germplasm to the OU PIO using the unique GID instead of the non-unique germplasm name

# Dependencies
none

# Testing
1. Import multiple germplasm with the same name
2. Import an experiment using GIDs from step 1.
3. After import successful, go to Experiment view and select the imported Experiment details
4. Verify that the correct GID is displayed.
See the card for sample import files.

# Checklist:

[BI-2028]: https://breedinginsight.atlassian.net/browse/BI-2028?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/org/breedinginsight/brapps/importer/services/processors/ExperimentProcessor.java
+++ b/src/main/java/org/breedinginsight/brapps/importer/services/processors/ExperimentProcessor.java
@@ -1299,8 +1299,7 @@ public class ExperimentProcessor implements Processor {
                                          .forEach(obsUnit -> obsUnit.getBrAPIObject()
                                                                     .setGermplasmDbId(germplasm.getGermplasmDbId()));
     }
-
-//obsUnit.getBrAPIObject().getAdditionalInfo().getAsJsonObject().get(BrAPIAdditionalInfoFields.GID).getAsString()
+    
     private void updateStudyDependencyValues(Map<Integer, PendingImport> mappedBrAPIImport, String programKey) {
         // update location DbIds in studies for all distinct locations
         mappedBrAPIImport.values()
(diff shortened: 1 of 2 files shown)


src/main/java/org/breedinginsight/brapps/importer/services/processors/ExperimentProcessor.java
```diff
@@ -1291,16 +1291,15 @@ private void updateStudyDbId(BrAPIStudy study, String programKey) {
...
                                          .stream()
-                                         .filter(obsUnit -> obsUnit.getBrAPIObject()
-                                                                   .getGermplasmName() != null &&
-                                                 obsUnit.getBrAPIObject()
-                                                        .getGermplasmName()
-                                                        .equals(germplasm.getGermplasmName()))
+                                         .filter(obsUnit -> germplasm.getAccessionNumber() != null &&
+                                                 germplasm.getAccessionNumber().equals(obsUnit
+                                                         .getBrAPIObject()
+                                                         .getAdditionalInfo().getAsJsonObject()
```
nickpalladino, line 1297: Do we need a migration to add gids to existing observation units additionalinfo?

dmeidlin: Most existing OUs have the correct gemplasm dbId assigned, so a migration adding the GID wouldn't matter either way for them. It's a question more for Alex and Shawn to estimate how many cases of existing OUs using germplasm there are. They haven't com across this until now, so I'm guessing it's a small number where manually fixing it in the species database would make more sense. I'm not sure what a migration would look like. It wouldn't be a case of just adding the GID to the OU additional info since the existing id in the OU germplasmDbId field could be wrong. You couldn't easily tell if it was right or wrong, and if it's wrong, it's not clear which germplasmDbId would be correct? So a migration just looks like re-importing the experiment with bug fix in place.

nickpalladino (approved):

davedrp (approved):
passed developer test

PR #331, opened 2024-02-14
BI-2053 - Fixed Scale Units Bug

mlm483:
# Description
**Story:** https://breedinginsight.atlassian.net/browse/BI-2053

### What was wrong:
The `ScaleEntity` class was missing a `units` field. The BrAPI `Scale` object has a `units` field, but our code was not previously using it. We were using `scaleName` to store units, but only numeric scale classes have units, and `scaleName` is a required BrAPI field, so we were just filling it with the scale class when units were not present.

### What this PR does:
- Our code now stores scale units in the `units` field of the BrAPI `Scale` and bi-api internal `ScaleEntity` objects.
- Since `scaleName` is required, and it doesn't actually exist in our Ontology template, I made the executive decision to set `scaleName` to `units` if provided, otherwise it falls back to the scale class (a.k.a. `dataType`). This logic is in **TraitFileParser.java**.
- The flyway (SQL) migration to add the `units` field to `ScaleEntity` also updates any existing numeric scales, setting `units` to the `scale_name`. I don't think this will do anything in production (we're wiping out brapi-server data), but it might help QA/rel-test go smoother.
- Per Shawn's guidance, attempting to upload an ontology file with Units specified with non-numeric scale classes now results in the tabular error message in [row #5 (as of writing) on this page](https://breedinginsight.atlassian.net/wiki/x/kIDEYw). 
- Updates integration/unit tests, including test ontology (trait) files. I added a test for the new validation.
- I made [a TAF PR](https://github.com/Breeding-Insight/taf/pull/137), it is untested, I recommend keeping it open until TAF is in better working order, I don't think it should hold up merging this PR.

> Note: if a Trait has the "Computation" method class, the scale class is ignored and overwritten with "Numerical", therefore units are allowed and indeed required. I don't know how well documented this is.

# Dependencies
bi-api: bug/BI-2053 branch
bi-web: release/v0.9 branch

# Testing
> **[Developers Only]** Because the schema changes, I had to run the following first, then run bi-api.
```shell
clean validate flyway:repair flyway:migrate install -D maven.test.skip=true --settings settings.xml
```

1. Try uploading [AlexC_Active_Ontology_ERROR.xls](https://github.com/Breeding-Insight/bi-api/files/14285271/AlexC_Active_Ontology_ERROR.xls) and confirm you see a tabular error message for the non-numerical rows with units.
2. Try uploading [AlexC_Active_Ontology_SUCCESS.xls](https://github.com/Breeding-Insight/bi-api/files/14285282/AlexC_Active_Ontology_SUCCESS.xls) and confirm the preview, list view, etc. all look OK.
3. Download the ontology uploaded in step 2. and confirm that only numerical rows have units.
4. Try uploading [Ont_Comp_SUCCESS.xlsx](https://github.com/Breeding-Insight/bi-api/files/14313835/Ont_Comp_SUCCESS.xlsx) and ensure that it succeeds and the "text" scale class (2nd row) is overwritten to be "Numerical".

# Checklist:


davedrp (approved):
Developer testing passed

nickpalladino (approved):
Don't remember why units wasn't used previously, maybe Breedbase didn't store them, but wouldn't really matter since stored in bi db.

mlm483:
> Don't remember why units wasn't used previously, maybe Breedbase didn't store them, but wouldn't really matter since stored in bi db.

I saw that in an older import template version there was a "Scale Name" column, that was dropped at some point and "Units" was added. 🤷

PR #360, opened 2024-05-23
[BI-2122-A] Workflows Select

dmeidlin:
# Description
**Story:** [BI-2122](https://breedinginsight.atlassian.net/browse/BI-2122?atlOrigin=eyJpIjoiZmI1ODZlOWU0ODVkNDdhOGE0ZGY0MzhiODU5ZTVmNDEiLCJwIjoiaiJ9)

Requiring system-generated uuids for workflows in import endpoints isn't really necessary or desirable as it spreads the maintaining of workflow changes across different parts of the application instead of encapsulating it according to the Stragey behavioral pattern. 

Each import service will only have a handful of workflows, so using appropriate human-readable identifier in the path of endpoints is preferable. The mapping id already identifies the import type, so workflow identifiers could even be reused without a naming collision.

- Workflow interface created having single method for processing the context
- ExperimentWorkflow, GermplasmWorkflow, SampleSubmissionWorkflow interfaces extend Workflow

## Navigator
A navigator is created that is injected as a dependency where ever a workflow must be selected for processing the context. The navigator encapsulates information only about the possible options for workflows (display name, path identifier) and can produce a list of all possible workflows.
- ExperimentWorkflowNavigator (navigator in Strategy pattern) implement ExperimentWorkflow

## Workflow
Workflows are created for processing the context according to a strategy. Each workflow encapsulates the information specific to that strategy (logic for selecting the workflow, the workflow order, and processing middleware).
- CreateNewExperimentWorkflow  implements ExperimentWorkflow
- AppendOverwritePhenotypesWorkflow implements ExperimentWorkflow
- CreateNewEnvironmentsWorkflow implements ExperimentWorkflow

##
- replaced system-generated workflow uuid with human-readable identifier in path of import endpoints
- deleted unnecessary factory, DAO, and db migration 

# Dependencies
- bi-web feature/BI-2121

# Testing
_Please include any details needed for reviewers to test this code_

# Checklist:

[BI-2122]: https://breedinginsight.atlassian.net/browse/BI-2122?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/main/java/org/breedinginsight/brapps/importer/model/workflow/ExperimentWorkflow.java
```diff
@@ -0,0 +1,6 @@
+package org.breedinginsight.brapps.importer.model.workflow;
+
+@FunctionalInterface
+public interface ExperimentWorkflow extends Workflow {
```
nickpalladino, line 4: Is it necessary to have to levels of Workflow interface implementations? Here and ExperimentWorkflowNavigator

nickpalladino (commented):
Here are my thoughts on the tradeoffs between the two different approaches:

- Database config vs in code config
    - Agree that config doesn't need to be in database but allows all config to be defined in one spot for any type of import
    - If we ever wanted to see all workflows in the system would be more straightforward to read entries out of import_mapping_workflow table (probably not too important)
- FileImportService level vs processor level construction
    - Getting workflow instance at the FileImportService level reduces the amount of downstream work required. Tradeoff is it's somewhat opaque.
    - Implementation of single Workflow interface vs implementation for each import type of workflow (ExperimentWorkflow, GermplasmWorkflow, etc.) and navigator for each workflow type (ExperimentWorkflowNavigator, etc.)
- Factory vs FunctionalInterface
    - I think I understand the motivation of having each Workflow running the same type of checks to see if they should process but it seems to be extra boiler plate and not sure it is easier to understand. I think a Factory is simpler in this case even though it has the tradeoff of having knowledge of all types of Workflows but it's getting that from the db so it's still dynamic.

src/main/java/org/breedinginsight/brapps/importer/services/FileImportService.java
```diff
@@ -429,18 +420,12 @@ public ImportUpload setDynamicColumns(ImportUpload newUpload, Table data, Import
...
             try {
-                Workflow workflow = null;
-                if (workflowId != null) {
-                    Optional<Workflow> optionalWorkflow = workflowFactory.getWorkflow(workflowId);
-                    workflow = optionalWorkflow.orElse(null);
-                }
-
                 ImportServiceContext context = ImportServiceContext.builder()
                         .workflow(workflow)
                         .brAPIImports(finalBrAPIImportList)
```
dmeidlin, line 431: A workflow isn't needed at the FileImportService level, which is why, after instantiating it here with the factory, the workflow had to be passed in by reference to where it is needed at the ExperimentImportService level. This is not desirable as client code receiving the ref has no guarantee that workflow has not been mutated. 

A less error-prone approach is to pass by value any workflow context and instantiate the workflow at the level where it is used, in this case the ExperimentImportService. It is an experiment import workflow instantiated within the ExperimentImportService; I'm not really seeing anything opaque about that.

src/main/java/org/breedinginsight/brapps/importer/services/FileImportService.java
```diff
@@ -586,8 +571,12 @@ public List<ImportMapping> getSystemMappingByName(String name) {
...
 
-    public List<ImportMappingWorkflow> getWorkflowsForSystemMapping(UUID mappingId) {
-        return importMappingWorkflowDAO.getWorkflowsByImportMappingId(mappingId);
+    public List<ImportWorkflow> getWorkflowsForSystemMapping(UUID mappingId) throws DoesNotExistException {
+        ImportMapping mappingConfig = importMappingDAO.getMapping(mappingId)
+                .orElseThrow(() -> new DoesNotExistException("Cannot find mapping config associated with upload."));
+        BrAPIImportService importService = configManager.getImportServiceById(mappingConfig.getImportTypeId())
+                .orElseThrow(() -> new DoesNotExistException("Config with that id does not exist"));
+        return importService.getWorkflows();
     }
```
dmeidlin, line 580: I agree that there probably won't be a use case for getting all workflows for all services, but it could be implemented readily by reading all mappingIds from the database and calling this method on each one in a loop.

dmeidlin:
> Here are my thoughts on the tradeoffs between the two different approaches:
> 
>     * Database config vs in code config
>       
>       * Agree that config doesn't need to be in database but allows all config to be defined in one spot for any type of import
>       * If we ever wanted to see all workflows in the system would be more straightforward to read entries out of import_mapping_workflow table (probably not too important)
>

dmeidlin:
> Here are my thoughts on the tradeoffs between the two different approaches:
> 
>     * Database config vs in code config
>       
>       * Agree that config doesn't need to be in database but allows all config to be defined in one spot for any type of import
>       * If we ever wanted to see all workflows in the system would be more straightforward to read entries out of import_mapping_workflow table (probably not too important)
> 
Agree that it is an unlikely use case, but can still be easily done. See my comment in code.

>     * Factory vs FunctionalInterface
>       
>       * I think I understand the motivation of having each Workflow running the same type of checks to see if they should process but it seems to be extra boiler plate and not sure it is easier to understand. I think a Factory is simpler in this case even though it has the tradeoff of having knowledge of all types of Workflows but it's getting that from the db so it's still dynamic.

The issue isn't whether to store workflow data in the db or in code(memory). It's why separate workflow data from the workflows in the first place. Doing so risks introducing six different ways for future us to introduce a bug when that can be avoided by encapsulating all workflow data in the workflow.

 If some workflow metadata (name, order, path identifier, etc.) is separated in a factory or db from the rest of the workflow business logic, then future us could update, delete, or add to the metadata without making the corresponding change to the workflow business logic. Future us could also update, delete, or add a workflow without corresponding changes to the metadata in the db or factory.

The approach of encapsulating the workflow data together eliminates the need for future us to maintain fidelity between different parts of the system and is, therefore, less error-prone. The logic for selecting a workflow is not additional boilerplate since it has to reside somewhere, and keeping it with the workflow will make it easy to maintain. 

Micronaut annotations can automagically do things that aren't readily apparent, and require a look at the docs if it's unfamiliar, but using them can simplify the code. In this case, the @FunctionalInterface and @Primary annotations reproduce the two parts(Context and Strategy) of the Strategy pattern nicely with the Navigator and Workflow classes. Any implementation of the Strategy pattern will have these two parts, so I don't see it as introducing complexity, in fact using these annotations is a slick way to do it. I also use it for trait validation.

dmeidlin:
> Here are my thoughts on the tradeoffs between the two different approaches:
> 
>     * FileImportService level vs processor level construction
>       
>       * Getting workflow instance at the FileImportService level reduces the amount of downstream work required. Tradeoff is it's somewhat opaque.
>       * Implementation of single Workflow interface vs implementation for each import type of workflow (ExperimentWorkflow, GermplasmWorkflow, etc.) and navigator for each workflow type (ExperimentWorkflowNavigator, etc.)
See my comment in the code for details. The downstream work here is the same. A workflow instance (ExperimentWorkflowNavigator) is created here; whereas a Workflow instance was created before. The difference is that now workflow context is passed by value until it is needed instead of instantiating when a Workflow is not needed and passing it by reference, a pattern we are trying to avoid if possible.

src/main/java/org/breedinginsight/brapps/importer/model/workflow/ExperimentWorkflow.java
dmeidlin: I implemented this for ExperimentImportService, so you can see the details there. The same can be done for Germplasm and SampleSubmission once WorkflowNavigators are created.

nickpalladino (approved):

PR #383, opened 2024-08-06
BI-2257 - Experimental Collaborator DAO and Services for experimental_program_user_role

mlm483:
# Description
**Story:** https://breedinginsight.atlassian.net/browse/BI-2257

- Implemented create, read and delete service (and when needed, DAO) methods for ExperimentalCollaborator. Keep in mind that the `ExperimentalCollaboratorService` and `ExperimentalCollaboratorDAO` are for interacting with the `experiment_program_user_role` table, they were named to make the purpose of this table obvious, rather than to make the connection to the table obvious.
- Updated the `ProgramUserService` and `ProgramUserDAO` methods to get a list of `ProgramUser` by role.
- Added tests for the new `ExperimentalCollaboratorService` methods.

> Note: An example of how to get the userId to pass down into the service method can be seen in `BreedingMethodController::createProgramBreedingMethod`. This is similar to how the `ExperimentalCollaboratorService` will ultimately be used from controller methods.

# Testing
There is nothing to test manually until controller endpoints are implemented.

# Checklist:

### DIFF
--- a/src/main/java/org/breedinginsight/daos/ExperimentalCollaboratorDAO.java
+++ b/src/main/java/org/breedinginsight/daos/ExperimentalCollaboratorDAO.java
@@ -19,11 +19,17 @@ package org.breedinginsight.daos;
 
 import lombok.extern.slf4j.Slf4j;
 import org.breedinginsight.dao.db.tables.daos.ExperimentProgramUserRoleDao;
+import org.breedinginsight.dao.db.tables.pojos.ExperimentProgramUserRoleEntity;
 import org.jooq.Configuration;
 import org.jooq.DSLContext;
 
 import javax.inject.Inject;
 import javax.inject.Singleton;
+
+import java.time.OffsetDateTime;
+import java.util.UUID;
+
+import static org.breedinginsight.dao.db.Tables.EXPERIMENT_PROGRAM_USER_ROLE;
 
 @Slf4j
 @Singleton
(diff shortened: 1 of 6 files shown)


src/main/java/org/breedinginsight/daos/ProgramUserDAO.java
```diff
@@ -194,4 +194,17 @@ public List<ProgramUser> parseRecords(Result<Record> records, BiUserTable create
...
     }
+
+    public List<ProgramUser> getProgramUsersByRole(UUID programId, UUID roleId) {
+        BiUserTable createdByUser = BI_USER.as("createdByUser");
+        BiUserTable updatedByUser = BI_USER.as("updatedByUser");
+
+        Result<Record> records = getProgramUsersQuery(createdByUser, updatedByUser)
+                .where(PROGRAM.ACTIVE.eq(true))
+                .and(PROGRAM.ID.eq(programId))
+                .and(ROLE.ID.eq(roleId))
```
nickpalladino, line 205: I think this should also have `.and(PROGRAM_USER_ROLE.ACTIVE.eq(true)` like in the `existsAndActive` method to make sure that the program user is still active in the program.

src/main/java/org/breedinginsight/services/ExperimentalCollaboratorService.java
```diff
@@ -31,4 +36,18 @@ public class ExperimentalCollaboratorService {
     public ExperimentalCollaboratorService(ExperimentalCollaboratorDAO experimentalCollaboratorDAO) {
         this.experimentalCollaboratorDAO = experimentalCollaboratorDAO;
     }
+
+    public ExperimentProgramUserRoleEntity createExperimentalCollaborator(UUID programUserRoleId, UUID experimentId, UUID userId) {
```
nickpalladino, line 40: Will there be a need to wrap `ExperimentProgramUserRoleEntity` into our own object for the purposes of returning via the api. I think sometimes we do that for serialization purposes, not sure if that will come up here.

src/main/java/org/breedinginsight/daos/ExperimentalCollaboratorDAO.java
```diff
@@ -36,4 +42,22 @@ public ExperimentalCollaboratorDAO(Configuration config, DSLContext dsl) {
         super(config);
         this.dsl = dsl;
     }
+
+    public ExperimentProgramUserRoleEntity create(UUID experimentId, UUID programUserRoleId, UUID userId) {
```
nickpalladino, line 46: Was the primary reason for doing it this way rather than using the DAO `insert` to have the data for returning without needing to do a select query after writing?

mlm483: Exactly.

src/main/java/org/breedinginsight/services/ExperimentalCollaboratorService.java
mlm483: The need could arise, I don't think we would wrap ExperimentProgramUserRoleEntity, but rather extend/inherit it and swapping in the subclass for the parent class shouldn't be a major effort.

As an example, I'm looking at `ProgramUpload<T>`, but that seems like a special case, a lot of our DAOs use the jooq generated entity (which all implement `Serializable` FWIW) such as `BreedingMethodDAOImpl` methods returning the jooq generated `ProgramBreedingMethodEntity`.

src/main/java/org/breedinginsight/daos/ProgramUserDAO.java
mlm483: Done.

src/main/java/org/breedinginsight/services/ExperimentalCollaboratorService.java
nickpalladino: Ya sorry intended to mean extend, not compose. Agree we can just deal with it if the need arises.

nickpalladino (approved):

davedrp (approved):

HMS17 (approved):

PR #392, opened 2024-08-21
[BI-2267] Trim trailing white space on breeding method file uploads

davedrp:
[BI-2267](https://breedinginsight.atlassian.net/browse/BI-2267) Trim trailing white space on breeding method file uploads

# Description
When importing germplasm, if there was a trailing blank in a _breeding method_ value, then the system was generating the table error  `Invalid breeding method`.

This fix removes any leading and trailing spaces in the imported _breeding method_ values.

# Dependencies
_Please include any dependencies to other code branches, testing configurations, scripts to be run, etc._

# Testing

1. Import a germplasm file with a trailing blanks in some of the _breeding method_ field.  You can use [ABC_germplasm_2024-08-02_06-20-46+0000import.xlsx](https://github.com/user-attachments/files/16697377/ABC_germplasm_2024-08-02_06-20-46%2B0000import.xlsx) which has trailing blanks in every _breeding method_ with the value "unknown "

**EXPECTED RESULT**
There should be no Table Error `Invalid breeding method`

# Checklist:

[BI-2267]: https://breedinginsight.atlassian.net/browse/BI-2267?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/org/breedinginsight/brapps/importer/services/processors/GermplasmProcessor.java
+++ b/src/main/java/org/breedinginsight/brapps/importer/services/processors/GermplasmProcessor.java
@@ -325,6 +325,7 @@ public class GermplasmProcessor implements Processor {
     private void processNewGermplasm(Germplasm germplasm, ValidationErrors validationErrors, Map<String, ProgramBreedingMethodEntity> breedingMethods,
                                      List<String> badBreedingMethods,
                                      Program program, UUID importListId, boolean commit, PendingImport mappedImportRow, int i, User user, Supplier<BigInteger> nextVal) {
+        germplasm = removeBreedingMethodBlanks(germplasm);
         // Get the breeding method database object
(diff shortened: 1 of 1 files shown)


mlm483 (approved):
I wonder if the meaning of `trimBreedingMethod` or `stripBreedingMethod` would be more clear than "remove...blanks" to developers. What do you think?

I think it was decided that the scope of this bug would remain narrow, so I'm fine with it only being for the breeding method. If we want to refactor the method to strip/trim more fields in the future it won't be hard to do. In that vein... you could name the method more generically like `stripWhitespace`. I'm happy either way.

dmeidlin (approved):

PR #402, opened 2024-09-18
BI-2308 - Field Book BrAPI Integration - Importing Fields

nickpalladino:
# Description
**Story:** [BI-2308](https://breedinginsight.atlassian.net/browse/BI-2308?atlOrigin=eyJpIjoiZWM1Y2RhZjAwMzMwNGNmN2ExZWIyNzg0OTQ0ZWMxOTciLCJwIjoiaiJ9)

- Branch name doesn't match but it's for BI-2308
- modified `removeProgramKeyAndUnknownAdditionalData` as regex seemed incorrect and was removing until the second ] and resulting pedigree string was incorrect. The method is used a bunch if different places so should make sure there aren't regressions. Not sure if issues have been noticed in the past?
- Ended up not removing setDbIds methods which was discussed as a possible approach in IT planning. The scope of doing that was much larger because the BrAPI objects in the cache are keyed off the DeltaBreed UUID, not the brapi dbId.

# Dependencies
- Field Book v5.6.24

# Testing
GIVEN have selected a field on the FieldBook Fields page to import
WHEN field is import is started
THEN field should successfully load in a time proportional to the amount of data, not hang indefinitely, and not get an error message

GIVEN have imported a Field with germplasm containing synonyms and pedigree
WHEN the attributes are viewed on the collect page
THEN should see synonyms and pedigree string with program keys & accession numbers stripped out

# Checklist:

[BI-2308]: https://breedinginsight.atlassian.net/browse/BI-2308?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/main/java/org/breedinginsight/brapi/v2/BrAPIGermplasmController.java
```diff
@@ -105,19 +113,88 @@ public HttpResponse<Object> searchGermplasm(
...
+        }
+        body.setGermplasmDbIds(convertedDbIds);
+
         ApiResponse<Pair<Optional<BrAPIGermplasmListResponse>, Optional<BrAPIAcceptedSearchResponse>>> brapiGermplasm;
         brapiGermplasm = brAPIEndpointProvider
                 .get(programDAO.getCoreClient(program.get().getId()), GermplasmApi.class)
                 .searchGermplasmPost(body);
 
+        return getObjectHttpResponse(brapiGermplasm, program.get().getKey());
+    }
```
dmeidlin, line 132:
```suggestion
        // convert request filter dbIds from DeltaBreed UUID to BrAPI service dbIds for now
        // TODO: all dbIds, just germplasmDbIds for now since that's what Field Book needs
        // Could be more performant by doing batch lookup but at least just going to cache
             body.setGermplasmDbIds(convertDbIdsToServiceIds(program.get().getId(), body.getGermplasmDbIds()));

        ApiResponse<Pair<Optional<BrAPIGermplasmListResponse>, Optional<BrAPIAcceptedSearchResponse>>> brapiGermplasm;
        brapiGermplasm = brAPIEndpointProvider
                .get(programDAO.getCoreClient(program.get().getId()), GermplasmApi.class)
                .searchGermplasmPost(body);

        return getObjectHttpResponse(brapiGermplasm, program.get().getKey());
    }

    private List<String> convertDbIdsToServiceIds(UUID programId, List<String> germplasmDbIds) {
        return germplasmDbIds.stream()
                .map(dbId -> {
                    try {
                        BrAPIGermplasm germplasm = germplasmService.getGermplasmByUUID(programId, dbId);
                        return germplasm.getGermplasmDbId();
                    } catch (Exception e) {
                        log.warn("Failed to convert dbId: " + dbId, e);
                        return null;
                    }
                })
                .filter(Objects::nonNull)
                .collect(Collectors.toList());
    }
```

src/main/java/org/breedinginsight/brapi/v2/BrAPIGermplasmController.java
```diff
@@ -105,19 +113,88 @@ public HttpResponse<Object> searchGermplasm(
...
+
+    private HttpResponse<Object> getObjectHttpResponse(ApiResponse<Pair<Optional<BrAPIGermplasmListResponse>, Optional<BrAPIAcceptedSearchResponse>>> brapiGermplasm,
+                                                       String programKey) throws ApiException {
         if (brapiGermplasm.getBody().getLeft().isPresent()) {
-            return HttpResponse.ok(brapiGermplasm.getBody().getLeft().get());
+            // convert dbIds to DeltaBreed UUID
+            BrAPIGermplasmListResponse response = brapiGermplasm.getBody().getLeft().get();
+            List<BrAPIGermplasm> germplasm = response.getResult().getData();
+            germplasm.forEach(g -> setDbIdsAndStripProgramKeys(g, programKey));
```
dmeidlin, line 166:
```suggestion
            batchProcessGermplasm(germplasm, programKey);
```

src/main/java/org/breedinginsight/brapi/v2/BrAPIGermplasmController.java
```diff
@@ -105,19 +113,88 @@ public HttpResponse<Object> searchGermplasm(
...
+            for (BrAPIGermplasmSynonyms synonym: germplasm.getSynonyms()) {
+                String newSynonym = Utilities.removeProgramKey(synonym.getSynonym(), programKey, germplasm.getAccessionNumber());
+                synonym.setSynonym(newSynonym);
+            }
+        }
 
+        // Remove program keys from pedigree string
+        String strippedPedigree = Utilities.removeProgramKeyAndUnknownAdditionalData(germplasm.getPedigree(), programKey);
+        germplasm.setPedigree(strippedPedigree);
     }
```
dmeidlin, line 198:
```suggestion
     private void batchProcessGermplasm(List<BrAPIGermplasm> germplasmList, String programKey) throws IllegalStateException {
        // Prepare a regex pattern for program key removal
        Pattern programKeyPattern = Pattern.compile(String.format("\\s*\\[%s-.*?\\]\\s*", programKey));

        germplasmList.parallelStream().forEach(germplasm -> {
            // Set dbId
            germplasm.germplasmDbId(Utilities.getExternalReference(germplasm.getExternalReferences(), "breedinginsight.org")
                    .orElseThrow(() -> new IllegalStateException("No BI external reference found"))
                    .getReferenceId());

            // Process synonyms
            if (germplasm.getSynonyms() != null) {
                germplasm.getSynonyms().forEach(synonym -> {
                    synonym.setSynonym(Utilities.removeProgramKey(synonym.getSynonym(), programKey, germplasm.getAccessionNumber()));
                });
            }

            // Process pedigree
            if (germplasm.getPedigree() != null) {
                germplasm.setPedigree(programKeyPattern.matcher(germplasm.getPedigree()).replaceAll(""));
            }
        });
    }
```

src/main/java/org/breedinginsight/utilities/Utilities.java
```diff
@@ -162,8 +162,21 @@ public static Object formatBrapiObjForDisplay(Object brapiInstance, Class brapiC
...
+     * .*?: Matches any character (except newline) zero or more times, non-greedily.
+     * . matches any character except newline.
+     * * means "zero or more times".
+     * ? makes the matching non-greedy, so it stops at the first closing bracket.
+     * \]: Matches the closing square bracket ]. Again, the backslash is used to escape it.
+     * \s*: Matches zero or more whitespace characters after the closing bracket.
+     * @param original
+     * @param programKey
+     * @return
+     */
```
dmeidlin, line 177: Unit tests for this method covering these cases with specific assertions for the expected format of synonyms and pedigree strings aster stripping would be helpful when testing for regressions later. New edge cases found could be added along with tests for very large fields or complex pedigrees.

dmeidlin (changes requested):
The setDbIdsAndStripProgramKeys method in BrAPIGermplasmController.java processes each germplasm item individually. For large datasets, this could potentially impact performance. I've added a method for processing a germplasm batch that only compiles the regex once and uses parallel streams if possible.

For clarity, I extracted the logic for replacing delta ids with dbIds into a private method to keep the body of the controller method lean.

I also think unit tests for `removeProgramKeyAndUnknownAdditionalData` are a good idea, but probably exceeds the effort level estimate for this card, but adding it to the card that checks for regressions in other places this method is used should include writing the tests.

src/main/java/org/breedinginsight/brapi/v2/BrAPIGermplasmController.java
nickpalladino: I'm good with breaking out a separate method but as far as the exception handling I think it'd be better to let that propagate up so we get an internal server error. I think it should be all or nothing and not tolerate any failure to convert ids.

src/main/java/org/breedinginsight/brapi/v2/BrAPIGermplasmController.java
nickpalladino: Were you advocating not changing `removeProgramKeyAndUnknownAdditionalData` in this card? I could see that so as to not take a chance breaking other stuff but if so should probably make that clear in the comments and maybe still put the regex string in a common area.

dmeidlin: Yep. The code suggestion above is just adding the batch method and using it to replace the for loop. I was not thinking of deleting `removeProgramKeyAndUnknownAdditionalData` for the reason you mentioned.

davedrp (approved):

mlm483 (commented):
I tested locally with Field Book version 5.6.24.

The first test passed, I was able to load fields at around 1000 germplasm/minute which seems fast enough for now.

The second test is less clear to me, I'm not sure where I'm meant to see synonym or pedigree info, but I didn't see any. The germplasm names in the collect screen didn't have the program key which is good.

Also, using wireshark, I see there was one request made from bi-api to brapi-server to the /brapi/v2/search/germplasm POST endpoint with an external reference and a list of germplasmDbIds. A get is then made for the search results (/brapi/v2/search/germplasm/{searchResultDbId}) but what's odd is the results are not ready when this request is made, and the client (bi-api) never seems to make a subsequent request, so I'm not sure what's happening there.

dmeidlin (changes requested):
Just re-submitting initial request

nickpalladino:
> I tested locally with Field Book version 5.6.24.
> 
> The first test passed, I was able to load fields at around 1000 germplasm/minute which seems fast enough for now.
> 
> The second test is less clear to me, I'm not sure where I'm meant to see synonym or pedigree info, but I didn't see any. The germplasm names in the collect screen didn't have the program key which is good.

If you included synonyms and pedigree information in the germplasm you should be able to tap the attributes on the collect page to open a dialog to increase the number of attributes shown and select the attributes you would like displayed. Synonyms and pedigree should be selectable options in the list.
> 
> Also, using wireshark, I see there was one request made from bi-api to brapi-server to the /brapi/v2/search/germplasm POST endpoint with an external reference and a list of germplasmDbIds. A get is then made for the search results (/brapi/v2/search/germplasm/{searchResultDbId}) but what's odd is the results are not ready when this request is made, and the client (bi-api) never seems to make a subsequent request, so I'm not sure what's happening there.

From what I was seeing the BrAPI server will give an immediate response if the number of dbIds being searched is below a certain threshold then switch to the searchResultDbId after that point. Field Book should make requests until it gets the result data.

mlm483:
With a small amount of data, I was able to see synonyms and pedigree. Using wireshark, I could see the response to the POST to the /brapi/v2/search/germplasm endpoint contained a searchResultDbId, and the data was returned in the response to the first GET request to the /brapi/v2/search/germplasm/{searchResultDbId} endpoint.

Screenshot 2024-10-01 at 9 44 35 AM


With a moderate amount of data (5000 Germplasm, ~1000 referenced by Observation Units), Field Book times out after 2 minutes when previewing the field (120 seconds is the default value for BRAPI_TIMEOUT in Field Book). The call that it times out on is the germplasm search request. Looking at the bi-api logs, I can see it's hitting the cache once for each Germplasm record, I think it's important to make that more efficient in this story.

Screenshot 2024-10-01 at 10 14 22 AM

src/main/java/org/breedinginsight/brapi/v2/BrAPIGermplasmController.java
nickpalladino: Ended up writing new service/DAO methods to convert all ids in one batch so could be done inline and a separate method is not longer needed.

src/main/java/org/breedinginsight/brapi/v2/BrAPIGermplasmController.java
nickpalladino: pushed changes

src/main/java/org/breedinginsight/brapi/v2/BrAPIGermplasmController.java
```diff
@@ -105,19 +113,88 @@ public HttpResponse<Object> searchGermplasm(
...
         body.externalReferenceIds(List.of(extRefId));
 
+        // convert request filter dbIds from DeltaBreed UUID to BrAPI service dbIds for now
+        // TODO: all dbIds, just germplasmDbIds for now since that's what Field Book needs
+        // Could be more performant by doing batch lookup but at least just going to cache
+        List<String> convertedDbIds = new ArrayList<>();
+        for (String germplasmDbId : body.getGermplasmDbIds()) {
+            BrAPIGermplasm germplasm = germplasmService.getGermplasmByUUID(program.get().getId(), germplasmDbId);
+            convertedDbIds.add(germplasm.getGermplasmDbId());
+        }
```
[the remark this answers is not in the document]
nickpalladino: Pushed new DAO and service methods

mlm483 (approved):
Other than the Field Book bug (I see your [issue](https://github.com/PhenoApps/Field-Book/issues/1055)), this is working as expected.

PR #411, opened 2024-10-09
[BI-2328] Experimental observation file import removes periods from column headers

davedrp:
[BI-2328](https://breedinginsight.atlassian.net/browse/BI-2328) Experimental observation file import removes periods from column headers

Added a validation so that a user can no longer add an ontology term with a period (.) in the Name field. 

# Dependencies
bi-web: **bug/BI-2328**

# Testing
**Test 1 (Manually add Ontology)**
1. Given user is logged in the application as a Program Administrator
2. Given user has selected an appropriate program (one in which they are a Program Administrator).
3. WHEN user selects "Ontology" in navigation
4. AND user selects 'New Term' button on ontology list page
5. AND user enters a text value with a period ('.') into the 'Name' field
6. AND user selects 'Save' button
7. THEN an error banner will appear with the text "Period in name is invalid"
8. AND the error message  "Period in name is invalid" will appear below the 'Name' field

**Test 2 (import Ontology)**
1. Given user is logged in the application as a Program Administrator
2. Given user has selected an appropriate program (one in which they are a Program Administrator).
3. WHEN user selects "Ontology" in navigation
4. AND user selects "Manage Ontology" button
5. AND user selects "Import file" link
6. AND user uploads Ontology "[bi_ontology_UAT-tyrwh-oat.xls](https://github.com/user-attachments/files/17317965/bi_ontology_UAT-tyrwh-oat.xls)" file
7. AND user selects 'Import' button
8. THEN an error banner will appear with the text, "`Error(s) detected in file, bi_ontology_UAT-tyrwh-oat.xls . (See details below.) Import cannot proceed`."
9. AND the error table will have the following values:
- Row = 7
- Field = 'Name'
- Error = 'Period is invalid'

# Checklist:

- [X] I have create/modified unit tests to cover this change
- [X] I have made corresponding changes to documentation

[BI-2328]: https://breedinginsight.atlassian.net/browse/BI-2328?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/org/breedinginsight/services/validators/TraitValidatorError.java
+++ b/src/main/java/org/breedinginsight/services/validators/TraitValidatorError.java
@@ -66,6 +66,10 @@ public class TraitValidatorError implements TraitValidatorErrorInterface {
     }
 
     @Override
+    public ValidationError getPeriodObsVarNameMsg() {
+        return new ValidationError("observationVariableName", "Period in name is invalid", HttpStatus.BAD_REQUEST);
+    }
+
     public ValidationError getMissingObsVarNameMsg() {
         return new ValidationError("observationVariableName", "Missing Name", HttpStatus.BAD_REQUEST);
     }
--- a/src/main/java/org/breedinginsight/services/validators/TraitValidatorService.java
+++ b/src/main/java/org/breedinginsight/services/validators/TraitValidatorService.java
@@ -16,6 +16,7 @@
  */
 package org.breedinginsight.services.validators;
 
+import io.micronaut.http.HttpStatus;
 import org.breedinginsight.api.model.v1.response.ValidationError;
 import org.breedinginsight.api.model.v1.response.ValidationErrors;
 import org.breedinginsight.dao.db.enums.DataType;
@@ -207,6 +208,23 @@ public class TraitValidatorService {
         return errors;
     }
 
+    public ValidationErrors checkTraitFieldsFormat(List<Trait> traits, TraitValidatorErrorInterface traitValidatorErrors) {
+
+        ValidationErrors errors = new ValidationErrors();
+
+        for (int i = 0; i < traits.size(); i++) {
+
+            Trait trait = traits.get(i);
+            String name = trait.getObservationVariableName();
+
+            if (name != null && name.contains(".")){
+                ValidationError error = traitValidatorErrors.getPeriodObsVarNameMsg();
+                errors.addError(traitValidatorErrors.getRowNumber(i), error);
+            }
+        }
+        return errors;
+    }
+
     public List<Trait> checkDuplicateTraitsExistingByName(UUID programId, List<Trait> traits){
 
         List<Trait> duplicates = new ArrayList<>();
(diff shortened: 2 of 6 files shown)


src/test/java/org/breedinginsight/model/delta/DeltaEntityFactoryUnitTest.java
```diff
@@ -426,7 +426,6 @@ private BrAPIObservation createBrAPIObservation() {
             .additionalInfo(toJsonObject(additionalInfoString))
             .collector("intern")
             .externalReferences(createExternalReferences())
-            .geoCoordinates(BrApiGeoJSON.builder().geometry(null).type("Feature").build())
```
nickpalladino, line 429: What was the reason this was removed?

nickpalladino (approved):

src/test/java/org/breedinginsight/model/delta/DeltaEntityFactoryUnitTest.java
davedrp: this was a problem I had locally.  It should not have been uploaded.  I will revert it.
-Thanks

src/main/java/org/breedinginsight/services/validators/TraitValidatorError.java
```diff
@@ -66,6 +66,10 @@ public ValidationError getMissingScaleDataTypeMsg() {
     }
 
     @Override
+    public ValidationError getPeriodObsVarNameMsg() {
+        return new ValidationError("observationVariableName", "Period in name is invalid", HttpStatus.BAD_REQUEST);
+    }
```
dmeidlin, line 71: I think it would be good to keep the error messages consistent between TraitValidatortError and TraitFileValidatorError, so this message could be changed to "Period is invalid".

src/main/java/org/breedinginsight/services/validators/TraitValidatorService.java
```diff
@@ -207,6 +208,23 @@ public ValidationErrors checkTraitFieldsLength(List<Trait> traits, TraitValidato
...
+
+        for (int i = 0; i < traits.size(); i++) {
+
+            Trait trait = traits.get(i);
+            String name = trait.getObservationVariableName();
+
+            if (name != null && name.contains(".")){
+                ValidationError error = traitValidatorErrors.getPeriodObsVarNameMsg();
+                errors.addError(traitValidatorErrors.getRowNumber(i), error);
+            }
```
dmeidlin, line 223: This check is fine as is, but since this validation step is checking the entire trait format, it might be nice to use a regex here so that it can be easily updated in the future to handle additional characters or edge cases.

```suggestion
            Pattern pattern = Pattern.compile("\\.");
            Matcher matcher = pattern.matcher(name);
            boolean containsInvalidCharacter = matcher.find();
            
            if (name != null && containsInvalidCharacter){
                ValidationError error = traitValidatorErrors.getPeriodObsVarNameMsg();
                errors.addError(traitValidatorErrors.getRowNumber(i), error);
            }
```

dmeidlin:
Looks good, and tests passed. Thanks for including the test import file!

src/main/java/org/breedinginsight/services/validators/TraitValidatorError.java
davedrp: This is a banner error message.  Table messages have a field the tells the user which field contains the error, banner messages do not.  So, it was decided that this message should contain the name of the filed.

dmeidlin (approved):

PR #419, opened 2024-10-28
BI-2344 - Incorrect Germplasm Export in Case of Duplicate Germplasm Names

mlm483:
# Description
**Story:** https://breedinginsight.atlassian.net/browse/BI-2344

I introduced this regression in [BI-2304](https://breedinginsight.atlassian.net/browse/BI-2304).

Because germplasm names are not guaranteed to be unique within a program or list, code that used germplasm name as a key for lookups had unexpected behavior in the case of multiple distinct germplasm in a list that shared the same name (different GIDs, same name is allowed).

The solution was to use the "{germplasmName} [{programKey}-{accessionNumber}]" format as the key for lookups instead, which is what is stored in the backing BrAPI service. Finally, the program key and accession number are stripped from the germplasm name when building the output file data (we don't want to display that to the user).

# Testing
1. Upload a germplasm file with duplicate germplasm names. Small example attached.
[germplasmDuplicateNames.xlsx](https://github.com/user-attachments/files/17559957/germplasmDuplicateNames.xlsx)
2. Download all germplasm for the program.
3. Ensure export file is ordered by GID and has no duplicate GIDs (though it has duplicate names).

The screenshot below shows an example upload file, erroneous output (release/1.0 branch) and correct output (bug/BI-2344 branch).
Screenshot 2024-10-29 at 12 09 20 PM

# Checklist:

[BI-2304]: https://breedinginsight.atlassian.net/browse/BI-2304?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/org/breedinginsight/brapi/v2/services/BrAPIGermplasmService.java
+++ b/src/main/java/org/breedinginsight/brapi/v2/services/BrAPIGermplasmService.java
@@ -117,7 +117,8 @@ public class BrAPIGermplasmService {
 
             HashMap<String, Object> row = new HashMap<>();
             row.put("GID", Integer.valueOf(germplasmEntry.getAccessionNumber()));
-            row.put("Germplasm Name", germplasmEntry.getGermplasmName());
+            // Strip programKey and accessionNumber from germplasmName for the file output.
+            row.put("Germplasm Name", Utilities.removeProgramKeyAnyAccession(germplasmEntry.getGermplasmName(), program.getKey()));
             row.put("Breeding Method", germplasmEntry.getAdditionalInfo().get(BrAPIAdditionalInfoFields.GERMPLASM_BREEDING_METHOD).getAsString());
             String source = germplasmEntry.getSeedSource();
             row.put("Source", source);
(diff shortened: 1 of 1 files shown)


src/main/java/org/breedinginsight/brapi/v2/services/BrAPIGermplasmService.java
```diff
@@ -82,10 +82,15 @@ public Optional<BrAPIGermplasm> getGermplasmByDBID(UUID programId, String germpl
...
-    public List<Map<String, Object>> processListData(List<BrAPIGermplasm> germplasm, BrAPIListDetails germplasmList){
+    public List<Map<String, Object>> processListData(List<BrAPIGermplasm> germplasm, BrAPIListDetails germplasmList, Program program){
         Map<String, BrAPIGermplasm> germplasmByName = new HashMap<>();
         for (BrAPIGermplasm g: germplasm) {
-            germplasmByName.put(g.getGermplasmName(), g);
+            // Use the full, unique germplasmName with programKey and accessionNumber (GID) for 2 reasons:
+            // 1. the BrAPI list items are full names, and
+            // 2. germplasmNames alone are not unique within a program, this led to unexpected behavior, see BI-2344.
+            String uniqueGermplasmName = String.format("%s [%s-%s]", g.getGermplasmName(), program.getKey(), g.getAccessionNumber());
+            g.setGermplasmName(uniqueGermplasmName);  // For later.
```
HMS17, line 92: could you clarify what the "for later" comment means?

HMS17 (approved):
Passed testing with multiple large germplasm lists with duplicates in a program

src/main/java/org/breedinginsight/brapi/v2/services/BrAPIGermplasmService.java
mlm483: I updated the comment to be more clear. Inside that loop, `g` is an element of the `germplasm` collection which is used again on line 111, and needs to have the full germplasmName with programKey and accession number later in the method.

davedrp (approved):
Lovely

PR #442, opened 2025-01-29
BI-2156-qa - Append Workflow Edge Cases

mlm483:
# Description
**Story:** https://breedinginsight.atlassian.net/browse/BI-2156 (These edge cases were discovered in QA while testing BI-2156, but they have nothing to do with BI-2156.)

### Current Behavior

When using the "Append experimental dataset" workflow (a) with a file that contains non-existent ObsUnitIds, and (b) with a file that contains valid ObsUnitIds but no Ontology Terms or Observation Data, generic errors were being shown to the user.

### Desired Behavior

Ideally (a) should result in an useful error message and (b) shouldn’t result in an error, but selecting commit shouldn’t change anything in the experiment.

### Changes
For (a) the main change was creating a better error message, `INVALID_OBS_UNIT_ID_ERROR` in `ExperimentUtilities.java`, the other changes (across ~~many~~ several files) were to propagate the exception correctly so that a 4xx status is returned rather than a 5xx status.

For (b) (commit https://github.com/Breeding-Insight/bi-api/pull/442/commits/e52481f645a2d1de0fee0ed8476b3efc617a24e8) I handled a null case to prevent a NullPointerException in `AppendOverwritePhenotypesWorkflow.java` at [this line](https://github.com/Breeding-Insight/bi-api/blob/110bdaa8ce8d704b194e7f0cb445f8a959675d04/src/main/java/org/breedinginsight/brapps/importer/services/processors/experiment/appendoverwrite/AppendOverwritePhenotypesWorkflow.java#L131).

# Testing

## (a)
1. Download an experiment from a different environment (e.g. qa-test) or different program, or create a new experiment file with non-existent ObsUnitIds.
3. Try to upload this file using the "Append" workflow.
4. You should see a useful error indicating the ObsUnitIds were not found.

## (b)
1. Upload an experiment with phenotypic data columns.
5. Download that experiment, and delete the phenotypic data columns from the file.
6. Re-upload the edited file using the "Append" workflow.
7. The upload should succeed without error and the experiment data should not be changed.

# Checklist:


dmeidlin (changes requested):
Getting and setting caught exceptions in the ImportWorkflowResult is deliberate and necessary for implementing a SAGA design pattern for handling transactions atomically. Changing the error messages and status codes is fine, but do not change the error handling to simply halt execution and propagate exceptions up the layers to be handled. This short-circuits the SAGA pattern and prevents compensating actions from being performed.

mlm483:
> Getting and setting caught exceptions in the ImportWorkflowResult is deliberate and necessary for implementing a SAGA design pattern for handling transactions atomically. Changing the error messages and status codes is fine, but do not change the error handling to simply halt execution and propagate exceptions up the layers to be handled. This short-circuits the SAGA pattern and prevents compensating actions from being performed.

I understand. I tried having `PendingObservationUnit::brapiRead` call `setProcessError` and then return an empty list in the case where invalid ObsUnitIds are supplied, but that's resulting in a NPE exception at a later step in the workflow. I'll see if I can get the workflow to run to completion anyways, it may be that other steps in the workflow could handle null or empty input more gracefully.

mlm483:
After discussions with Nick, Dave, and Shawn, I've made several changes.

1. Switched to using `ValidationErrors` to build a detailed tabular error which the frontend will display to the user.
2. Created `MissingValuesException` as a way to efficiently pass a set of strings up the call chain.
3. Updated the error message to match Shawn's documentation, https://cornell.box.com/s/67i3q3d3jkw5zenxngqufm4ft0boi3mf.

dmeidlin:
Caught exceptions are handled in the middleware instead of the lower layers, so just pass the exception through PendingObservationUnit#brapiRead and move the ValidationError handler up to the middleware.

dmeidlin:
The first middleware step in the appendOverwrite workflow is [AppendOverwriteIDValidation.java](https://github.com/Breeding-Insight/bi-api/blob/develop/src/main/java/org/breedinginsight/brapps/importer/services/processors/experiment/appendoverwrite/middleware/AppendOverwriteIDValidation.java) which currently checks for empty values or duplicated values in the ObsUnitID column, throwing an exception if any are found which appears as a banner message on the front end. Validating that every valid ObSUnitID value is also found in the BrAPI service should be done in this step. This involves simply moving [this line](https://github.com/Breeding-Insight/bi-api/blob/develop/src/main/java/org/breedinginsight/brapps/importer/services/processors/experiment/appendoverwrite/middleware/initialize/WorkflowInitialization.java#L63) from `WorkflowInitialization.java` to `AppendOverwriteIDValidation.java`, and placing the ValidationError handler here.

dmeidlin:
`MissingValuesExcepton` is a confusing name because there is already a check for missing ObsUnitID values being done, and this error type is not used for that case. The error type is really for the case where there isn't a BrAPI entity stored for the given id, so a different name like `EntityNotFoundException` would be appropriate. Also, instead of placing the class for this exception type at the root of the experiment package, `experiment/model` seems like a better place for it.

dmeidlin:
[`ExperimentUtilities#collateReferenceOUIds`](https://github.com/Breeding-Insight/bi-api/blob/110bdaa8ce8d704b194e7f0cb445f8a959675d04/src/main/java/org/breedinginsight/brapps/importer/services/processors/experiment/ExperimentUtilities.java#L304-L335) is the method currently called to both validate against missing or duplicated ids and collect all the ids if all rows pass validation. This method should be split into two methods, `ValidateReferenceOUIdValues` and `collateUniqueOUIds`. This way validation errors can be generated for missing or duplicated values and displayed in the preview table instead of just a banner message. Any remaining unique OU ids can then be collected and corresponding Obs Units initialized. if an `EntityNotFoundException` is thrown then the exception propagates up to the AppendOverwriteIDValidation middleware where validation errors are built and added to the others by calling `ExperimentUtilities#addValidationErrorsForObsUnitsNotFound`, which contains your handler logic.

dmeidlin:
There's a [bug](https://github.com/Breeding-Insight/bi-api/blob/67421212396716bfdc9abb015b27ff4306ecda79/src/main/java/org/breedinginsight/brapps/importer/services/processors/experiment/service/ObservationUnitService.java#L72-L73) in `ObservationUnitService#getObservationUnitsByDbId`. This method is fetching the OUs for a collection of Deltabreed ids, not for a collection of BrAPI Service DbIds. This method needs to search the xrefs, not the dbIds, of the OUs returned from the service. The method should also be renamed to `ObservationUnitService#getObservationUnitsById`.

dmeidlin:
I spoke with @shawnyarnes about the error messaging for the case of missing or duplicated ObsUnitID values and the source of truth has been updated and should be used for the validation errors.

dmeidlin:
There are a couple of problems with setting all summary statistics counts to zero in `ImportTableProcess` [here](https://github.com/Breeding-Insight/bi-api/blob/67421212396716bfdc9abb015b27ff4306ecda79/src/main/java/org/breedinginsight/brapps/importer/services/processors/experiment/appendoverwrite/middleware/process/ImportTableProcess.java#L389-L392).
The edge case being addressed is where an experiment exists without a dataset defined (no observation variable, no observation data) and the user selects the append workflow but imports a sheet with no new data (still no obs vars and no data). There isn't a check in place for this edge case, and the action taken is not correct for this edge case. The experiment still has germplasm and study counts that should appear in the summary preview, but are incorrectly being displayed as zero everywhere.
undefinedDatasetNoSummary
The expected summary for this edge case should look like the image below, instead.
undefinedDatasetWithSummary

Each import row and each cell of observation data can be in one of many possible states that requires its own set of validation steps, pending object processing, and statistics calculation, and there is a `ProcessedDataFactory` for creating instances of each data class containing the specific methods for validation, processing, statistics, etc., so that there isn't a confusing morass of nested if-else blocks and switch cases in the main import table process method. Instead of tacking on a bit of code to `ImportTableProcess#process` that sets the statistics to empty values(which isn't correct), you'll need to create a new class `appendoverwrite/factory/data/UndefinedDataset.java` and update [ProcessedDataFactory](https://github.com/Breeding-Insight/bi-api/blob/67421212396716bfdc9abb015b27ff4306ecda79/src/main/java/org/breedinginsight/brapps/importer/services/processors/experiment/appendoverwrite/factory/data/ProcessedDataFactory.java) to produce beans of that type.

Then in the import process method [just before looping over any observation variables defined](https://github.com/Breeding-Insight/bi-api/blob/67421212396716bfdc9abb015b27ff4306ecda79/src/main/java/org/breedinginsight/brapps/importer/services/processors/experiment/appendoverwrite/middleware/process/ImportTableProcess.java#L242-L244) add a check for this edge case where there are no observation variables and set `processedData` to a new instance of `UndefinedDatasetBean` using the factory. 

Then have processedData call [the same methods to update statistics](https://github.com/Breeding-Insight/bi-api/blob/67421212396716bfdc9abb015b27ff4306ecda79/src/main/java/org/breedinginsight/brapps/importer/services/processors/experiment/appendoverwrite/middleware/process/ImportTableProcess.java#L358-L365) as is used when processing rows with observation variables. It would be good to place these lines in a helper method `updatePreviewStatistics`.

src/main/java/org/breedinginsight/brapps/importer/services/processors/experiment/ExperimentUtilities.java
```diff
@@ -70,8 +70,7 @@ public class ExperimentUtilities {
     public static final String PREEXISTING_EXPERIMENT_TITLE = "Experiment Title already exists";
     public static final String MISSING_OBS_UNIT_ID_ERROR = "Experimental entities are missing ObsUnitIDs";
     public static final String UNMATCHED_COLUMN = "Ontology term(s) not found: ";
-
-
+    public static final String INVALID_OBS_UNIT_ID_ERROR = "Invalid ObsUnitID";
```
dmeidlin, line 73: The message for empty ids is the same but duplicate ids have a more specific error message. Also, constants specific to the append workflow are [here](https://github.com/Breeding-Insight/bi-api/blob/develop/src/main/java/org/breedinginsight/brapps/importer/services/processors/experiment/model/ExpImportProcessConstants.java).
```suggestion
        public enum ErrMessage {
        MULTIPLE_EXP_TITLES("File contains more than one Experiment Title"),
        MISSING_OBS_UNIT_ID("Invalid ObsUnitID"),
        PREEXISTING_EXPERIMENT_TITLE("Experiment Title already exists"),
        UNMATCHED_COLUMN("Ontology term(s) not found: "),
        OBS_UNIT_NOT_FOUND("Invalid ObsUnitID"),
        DUPLICATE_OBS_UNIT_ID("ObsUnitId is repeated");
```

src/main/java/org/breedinginsight/brapps/importer/services/processors/experiment/MissingValuesException.java
```diff
@@ -0,0 +1,12 @@
...
+import lombok.Getter;
+
+import java.util.Set;
+
+@Getter
+public class MissingValuesException extends Exception {
+    private Set<String> missingIds;
+
+    public MissingValuesException(Set<String> missingIds) { this.missingIds = missingIds; }
+}
```
dmeidlin, line 12:
```suggestion
package org.breedinginsight.brapps.importer.services.processors.experiment.model;

import lombok.Getter;

import java.util.Set;
@Getter
public class EntityNotFoundException extends Throwable {
    private Set<String> missingEntityIds;

    public EntityNotFoundException(Set<String> missingEntityIds) { this.missingEntityIds = missingEntityIds; }
}
```

src/main/java/org/breedinginsight/brapps/importer/services/processors/experiment/appendoverwrite/factory/action/BrAPIAction.java
```diff
@@ -42,7 +43,7 @@ public interface BrAPIAction<T> {
      * @return An Optional containing the relevant BrAPI state after executing the action.
      * @throws ApiException if an error occurs during the execution of the action.
      */
-    Optional<BrAPIState<T>> execute() throws ApiException, MissingRequiredInfoException, UnprocessableEntityException, DoesNotExistException;
+    Optional<BrAPIState<T>> execute() throws ApiException, MissingRequiredInfoException, UnprocessableEntityException, DoesNotExistException, ValidatorException;
```
dmeidlin, line 46:
```suggestion
Optional<BrAPIState<T>> execute() throws ApiException, MissingRequiredInfoException, UnprocessableEntityException, DoesNotExistException, EntityNotFoundException;
```

src/main/java/org/breedinginsight/brapps/importer/services/processors/experiment/appendoverwrite/factory/action/WorkflowReadInitialization.java
```diff
@@ -44,7 +45,7 @@ protected WorkflowReadInitialization(ExperimentImportEntity<T> entity) {
      * @return an Optional containing the BrAPIState representing the completed read workflow
      * @throws ApiException if an error occurs during execution
      */
-    public Optional<BrAPIState<T>> execute() throws ApiException {
+    public Optional<BrAPIState<T>> execute() throws ApiException, ValidatorException {
```
dmeidlin, line 48:
```suggestion
    public Optional<BrAPIState<T>> execute() throws ApiException, EntityNotFoundException {
```

src/main/java/org/breedinginsight/brapps/importer/services/processors/experiment/appendoverwrite/factory/entity/ExperimentImportEntity.java
```diff
@@ -43,7 +44,7 @@ public interface ExperimentImportEntity<T> {
      * @return List of fetched entities
      * @throws ApiException if there is an issue with the API call
      */
-    public List<T> brapiRead() throws ApiException;
+    public List<T> brapiRead() throws ApiException, ValidatorException;
```
dmeidlin, line 47:
```suggestion
    public List<T> brapiRead() throws ApiException, EntityNotFoundException;
```

src/main/java/org/breedinginsight/brapps/importer/services/processors/experiment/appendoverwrite/factory/entity/PendingObservationUnit.java
```diff
@@ -94,12 +98,26 @@ public List<BrAPIObservationUnit> brapiPost(List<BrAPIObservationUnit> members)
...
+
+            // Build a detailed tabular error.
+            for (int rowNum = 0; rowNum < importContext.getImportRows().size(); rowNum++) {
+                String rowObsUnitId = ((ExperimentObservation)importContext.getImportRows().get(rowNum)).getObsUnitID();
+                if (e.getMissingIds().contains(rowObsUnitId)) {
+                    ExperimentUtilities.addRowError(ExperimentObservation.Columns.OBS_UNIT_ID, ExperimentUtilities.INVALID_OBS_UNIT_ID_ERROR, validationErrors, rowNum);
+                }
+            }
+            throw new ValidatorException(validationErrors);
+        }
```
dmeidlin, line 120:
```suggestion
        @Override
    public List<BrAPIObservationUnit> brapiRead() throws ApiException, EntityNotFoundException {
        // Collect deltabreed-generated obs unit ids listed in the import
        Set<String> obsUnitIds = cache.getReferenceOUIds();

        // For each id fetch the observation unit from the brapi data store
        return observationUnitService.getObservationUnitsByDbId(new HashSet<>(obsUnitIds), importContext.getProgram());
    }
```

src/main/java/org/breedinginsight/brapps/importer/services/processors/experiment/appendoverwrite/middleware/initialize/WorkflowInitialization.java
```diff
@@ -66,7 +67,7 @@ public AppendOverwriteMiddlewareContext process(AppendOverwriteMiddlewareContext
             locationReadWorkflowInitialization.execute();
             brAPIDatasetReadWorkflowInitialization.execute();
             brAPIGermplasmReadWorkflowInitialization.execute();
-        } catch (ApiException e) {
+        } catch (ApiException | ValidatorException e) {
```
dmeidlin, line 70: Remove the OU initialization from this middleware and add it to the prior validation middleware step.
```suggestion
     @Override
    public AppendOverwriteMiddlewareContext process(AppendOverwriteMiddlewareContext context) {
        brAPITrialReadWorkflowInitialization = brAPIReadFactory.trialWorkflowReadInitializationBean(context);
        brAPIStudyReadWorkflowInitialization = brAPIReadFactory.studyWorkflowReadInitializationBean(context);
        locationReadWorkflowInitialization = brAPIReadFactory.locationWorkflowReadInitializationBean(context);
        brAPIDatasetReadWorkflowInitialization = brAPIReadFactory.datasetWorkflowReadInitializationBean(context);
        brAPIGermplasmReadWorkflowInitialization = brAPIReadFactory.germplasmWorkflowReadInitializationBean(context);

        log.debug("reading required BrAPI data from BrAPI service");
        try {
            brAPITrialReadWorkflowInitialization.execute();
            brAPIStudyReadWorkflowInitialization.execute();
            locationReadWorkflowInitialization.execute();
            brAPIDatasetReadWorkflowInitialization.execute();
            brAPIGermplasmReadWorkflowInitialization.execute();
        } catch (ApiException e) {
            context.getAppendOverwriteWorkflowContext().setProcessError(new MiddlewareException(e));
            return this.compensate(context);
        } catch (EntityNotFoundException e) {
            // TODO: handle edge cases of missing brapi entities as needed
            return this.compensate(context);
        }

        return processNext(context);
    }
```

src/main/java/org/breedinginsight/brapps/importer/services/processors/experiment/appendoverwrite/middleware/process/ImportTableProcess.java
```diff
@@ -386,6 +386,11 @@ public AppendOverwriteMiddlewareContext process(AppendOverwriteMiddlewareContext
             // Add the pending observation map to the context for use in processing the import
             context.getAppendOverwriteWorkflowContext().setPendingObservationByHash(pendingObservationByHash);
 
+            // Make sure the workflow statistic is not null.
+            if (context.getAppendOverwriteWorkflowContext().getStatistic() == null) {
+                context.getAppendOverwriteWorkflowContext().setStatistic(statistic);
+            }
+
```
dmeidlin, line 393:
```suggestion
                         /**
                 * Handle the edge case where a user imports with the append/overwrite workflow for an experiment
                 * without a dataset defined (i.e. no observation variables headers) and the import does not
                 * actually have new data to append
                 */
                if (phenotypeCols.isEmpty()) {
                    processedData = processedDataFactory.undefinedDatasetBean();
                    updatePreviewStatistics(processedData, context, studyName, unitId);
                }

                // Assemble the pending observation data for all phenotypes
                for (Column<?> column : phenotypeCols) {
                    String cellData = column.getString(rowNum);

```

src/main/java/org/breedinginsight/brapps/importer/services/processors/experiment/service/ObservationUnitService.java
```diff
@@ -61,7 +59,7 @@ public ObservationUnitService(BrAPIObservationUnitDAO brAPIObservationUnitDAO) {
      * @throws ApiException if an error occurs during the retrieval of observation units
      * @throws IllegalStateException if the retrieved observation units do not match the provided observation unit IDs
      */
-    public List<BrAPIObservationUnit> getObservationUnitsByDbId(Set<String> obsUnitIds, Program program) throws ApiException, IllegalStateException {
+    public List<BrAPIObservationUnit> getObservationUnitsByDbId(Set<String> obsUnitIds, Program program) throws ApiException, IllegalStateException, MissingValuesException {
```
dmeidlin, line 62: The method name should be changed, the bug fixed, and EntityNotFoundException used as a throwable.
```suggestion
        public List<BrAPIObservationUnit> getObservationUnitsById(Set<String> obsUnitIds, Program program) throws ApiException, IllegalStateException, EntityNotFoundException {
        List<BrAPIObservationUnit> brapiUnits = null;

        // Retrieve reference Observation Units based on IDs
        brapiUnits = brAPIObservationUnitDAO.getObservationUnitsById(obsUnitIds, program);

        // If no BrAPI units are found, throw an EntityNotFoundException with an error message
        if (obsUnitIds.size() != brapiUnits.size()) {
            Set<String> missingIds = new HashSet<>(obsUnitIds);

            // Calculate missing IDs based on retrieved BrAPI units
            //missingIds.removeAll(brapiUnits.stream().map(BrAPIObservationUnit::getObservationUnitDbId).collect(Collectors.toSet()));
            missingIds.removeAll(brapiUnits.stream()
                    .map(unit -> Utilities.getExternalReference(unit.getExternalReferences(), BRAPI_REFERENCE_SOURCE, ExternalReferenceSource.OBSERVATION_UNITS))
                    .filter(Optional::isPresent)
                    .map(Optional::get)
                    .map(BrAPIExternalReference::getReferenceId).collect(Collectors.toSet()));

            // Throw exception with missing IDs information
            throw new EntityNotFoundException(missingIds);
        }

        return brapiUnits;
    }
```

nickpalladino (approved):

dmeidlin (approved):

PR #444, opened 2025-02-17
[BI-2450]BrAPI Server DB connections exhausted on biapi startup caching

davedrp:
[BI-2450](https://breedinginsight.atlassian.net/browse/BI-2450) - BrAPI Server DB connections exhausted on biapi startup caching

# Description
The scheduled delay in creating the cache from the  DOA's is now driven by the .env file.  That way the expensive calls to the BrAPI server (that are called when bi-api is first started). can be staggered.  The timing can be changed in a config file (.env), so that it can be fine tuned in each environment.  

# Testing

- If the new parameters are NOT added to your local .env-file then there should be no change to the system's behavior.
- If the new parameters are added, then the caching should be scheduled accordingly.  
  - FOR EXAMPLE:
    - if the parameters are:
    - `germplasm_start_delay=5s
        study_start_delay=10s
         trial_start_delay=15s
         trait_start_delay=20s
         observation_start_delay=25s
         observation_unit_start_delay=30s`
       - THEN you should see line in you bi-api log that look something like:
       - 
 `2025-02-17 15:16:29.987-0500 [scheduled-executor-thread-1] DEBUG o.b.brapi.v2.dao.BrAPIGermplasmDAO - populating germplasm cache`

`2025-02-17 15:16:35.559-0500 [scheduled-executor-thread-3] DEBUG o.b.brapi.v2.dao.BrAPIStudyDAO - populating study cache`

`2025-02-17 15:16:40.548-0500 [scheduled-executor-thread-7] DEBUG o.b.b.v2.dao.impl.BrAPITrialDAOImpl - populating experiment cache`

`2025-02-17 15:16:50.548-0500 [scheduled-executor-thread-6] DEBUG o.b.brapi.v2.dao.BrAPIObservationDAO - populating observation cache`

`2025-02-17 15:16:55.551-0500 [scheduled-executor-thread-8] DEBUG o.b.b.v2.dao.BrAPIObservationUnitDAO - populating observation unit cache`

Note:  there is at least 5-seconds between each line.

# Checklist:

- [ x I have tested my code and ensured it meets the acceptance criteria of the story

[BI-2450]: https://breedinginsight.atlassian.net/browse/BI-2450?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -64,9 +64,9 @@ BRAPI_VENDOR_SUBMISSION_ENABLED=false #can a submission be sent to a vendor via 
 BRAPI_VENDOR_CHECK_FREQUENCY=1d #how often to check for vendor updates for sample submissions
 
 #The initial caching of each type of object needs to be staggered by the prescribed number of seconds
-germplasm_start_delay=5s
-study_start_delay=10s
-trial_start_delay=15s
-trait_start_delay=20s
-observation_start_delay=25s
-observation_unit_start_delay=30s
+GERMPLASM_START_DELAY=5s
+STUDY_START_DELAY=10s
+TRIAL_START_DELAY=15s
+TRAIT_START_DELAY=20s
+OBSERVATION_START_DELAY=25s
+OBSERVATION_UNIT_START_DELAY=30s
--- a/.env.template
+++ b/.env.template
@@ -62,3 +62,11 @@ AWS_S3_ENDPOINT=<s3 endpoint, default https://s3.us-east-1.amazonaws.com>
 
 BRAPI_VENDOR_SUBMISSION_ENABLED=false #can a submission be sent to a vendor via BrAPI
 BRAPI_VENDOR_CHECK_FREQUENCY=1d #how often to check for vendor updates for sample submissions
+
+#The initial caching of each type of object needs to be staggered by the prescribed number of seconds
+germplasm_start_delay=5s
+study_start_delay=10s
+trial_start_delay=15s
+trait_start_delay=20s
+observation_start_delay=25s
+observation_unit_start_delay=30s
--- a/src/main/resources/application.yml
+++ b/src/main/resources/application.yml
@@ -205,3 +205,11 @@ aws:
     buckets:
       genotype:
         bucket: ${AWS_GENO_BUCKET}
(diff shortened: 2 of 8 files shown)


.env.template
```diff
@@ -61,4 +61,12 @@ AWS_GENO_BUCKET=<s3 bucket for genotypic data uploads>
 AWS_S3_ENDPOINT=<s3 endpoint, default https://s3.us-east-1.amazonaws.com>
 
 BRAPI_VENDOR_SUBMISSION_ENABLED=false #can a submission be sent to a vendor via BrAPI
-BRAPI_VENDOR_CHECK_FREQUENCY=1d #how often to check for vendor updates for sample submissions
\ No newline at end of file
+BRAPI_VENDOR_CHECK_FREQUENCY=1d #how often to check for vendor updates for sample submissions
+
+#The initial caching of each type of object needs to be staggered by the prescribed number of seconds
```
nickpalladino, line 66: Could you make these new envs all caps to be consistent with the others?

src/main/resources/application.yml
```diff
@@ -205,3 +205,11 @@ aws:
     buckets:
       genotype:
         bucket: ${AWS_GENO_BUCKET}
+startup:
+  delay:
+    germplasm: ${germplasm_start_delay:2s}
```
nickpalladino, line 210: Can update these to caps as well when changed in env file

nickpalladino (changes requested):
I'm good with this approach as a first step. If we decide we want to simplify the config file in the future after doing some experimenting with the parameters we can always do that. Just had a comment about matching case of the new envs.

HMS17 (approved):

.env.template
davedrp: updated to ALL CAPS

src/main/resources/application.yml
davedrp: updated to ALL CAPS

nickpalladino (approved):

PR #463, opened 2025-05-08
[BI-2632] Germplasm download from single list no longer working

davedrp:
[BI-2632](https://breedinginsight.atlassian.net/browse/BI-2632) Germplasm download from single list no longer working

# Description
Attempting to download a Germplasm List results in a 403 error.

_Please include a summary of the change that was made._
The endpoint  @Get("/programs/{programId}/germplasm/lists/{listDbId}/export{?fileExtension}") was removed,  but is still being called.  The endpoint was restored.

# Dependencies
bi-web: release/1.1

# Testing
_Please include any details needed for reviewers to test this code_
- Select Germplasm from the side bar menu.
- Select the Lists tab.
-  Select Download for any of the Lists (on the right hand side of the table).
-  Select any file format from the modal screen.
-  Press the Download button.

EXPECTED RESULT
    File downloads successfully.

# Checklist:

- [x] I have performed a self-review of my own code
- [x] I have tested my code and ensured it meets the acceptance criteria of the story
- [ ] I have tested that my code works with both the brapi-java-server and BreedBase
- [ ] I have create/modified unit tests to cover this change
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to documentation
- [ ] I have run TAF: _\<please include a link to TAF run>_

[BI-2632]: https://breedinginsight.atlassian.net/browse/BI-2632?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #471, opened 2025-06-11
Bug/bi 2632

davedrp:
same as https://github.com/Breeding-Insight/bi-api/pull/463/files but merged off of _release/1.1.1_


davedrp:
Was already approved in https://github.com/Breeding-Insight/bi-api/pull/463, so approval is bypassed here.

PR #487, opened 2025-08-19
[BI-2193] Improve/Correct Sub-Entity Tabular View

HMS17:
# Description
**Story:** [BI-2193 - Improve/Correct Sub-Entity Tabular View](https://breedinginsight.atlassian.net/browse/BI-2193)

For ease of retrieving and displaying top level exp unit id in the sub-entity dataset view, the top level is exp unit id is now stored on creation of sub entity observation units in the additional info.

# Dependencies
bi-web: [feature/BI-2193](https://github.com/Breeding-Insight/bi-web/pull/450)

# Testing
see bi-web

# Checklist:


mlm483 (approved):

src/main/java/org/breedinginsight/brapi/v2/services/BrAPITrialService.java
```diff
@@ -573,6 +573,8 @@ public BrAPIObservationUnit createSubObservationUnit(
         levelRelationships.add(expUnitLevel);
         position.setObservationLevelRelationships(levelRelationships);
 
+        observationUnit.putAdditionalInfoItem(BrAPIAdditionalInfoFields.EXP_UNIT_ID, expUnit.getObservationUnitName());
+
```
nickpalladino, line 577: I think it may be worth standardizing how the level relationships are stored for the ExpUnit and SubObsUnit. The subObsUnit level code is set like this:

`level.setLevelCode(Utilities.appendProgramKey(subUnitId, program.getKey(), seqVal));`

The ExpUnit level is set like this:

`expUnitLevel.setLevelCode(Utilities.appendProgramKey(expUnitUUID, program.getKey(), seqVal));`

If they both stored the unitID in the level code I think you could use that instead of adding an additionalInfo field for it. It would be slightly more difficult to work with, wondering your thoughts?

nickpalladino: ok merging this and moving work related to comment to a new card

nickpalladino (approved):

PR #490, opened 2025-12-12
[BI-2234] - Improve exp dataset export file

HMS17:
# Description
**Story:** [BI-2234 - Improve exp dataset export file](https://breedinginsight.atlassian.net/browse/BI-2234)

Changes to make sub entity dataset export file contain desired columns while not altering the top level export file

- Added backend logic to distinguish between top level and sub entity datasets when exporting and change columns present accordingly
- Updated backend to fill additional sub-entity columns with correct data
- Updated sorting of export file in sub-entity case to also sort by Sub Unit ID (sorting of sub entity dataset UI will be in BI-2755)

# Dependencies
n/a

# Testing
- Create an experiment with both a top level and sub entity dataset
- Download top level dataset: check that file is sorted env > exp unit id, data is correct, and there are no sub-entity columns
- Download sub entity dataset: check that file is sorted env > exp unit id > sub unit id, data is correct, and there are the following columns present
-- Exp Unit
-- Sub-Obs Unit
-- Exp Unit ID
-- Sub Unit ID
-- (Exp Unit Level)  ObsUnitID
-- (Sub-Obs Unit Level) ObsUnitID

# Checklist:

### DIFF
--- a/src/main/java/org/breedinginsight/brapi/v2/services/BrAPITrialService.java
+++ b/src/main/java/org/breedinginsight/brapi/v2/services/BrAPITrialService.java
@@ -36,10 +36,7 @@ import org.breedinginsight.model.*;
 import org.breedinginsight.services.TraitService;
 import org.breedinginsight.services.exceptions.DoesNotExistException;
 import org.breedinginsight.services.parsers.experiment.ExperimentFileColumns;
-import org.breedinginsight.utilities.DatasetUtil;
-import org.breedinginsight.utilities.IntOrderComparator;
-import org.breedinginsight.utilities.FileUtil;
-import org.breedinginsight.utilities.Utilities;
+import org.breedinginsight.utilities.*;
 import org.jetbrains.annotations.NotNull;
 
 import javax.inject.Inject;
@@ -154,9 +151,6 @@ public class BrAPITrialService {
                 new ArrayList<>(Arrays.asList(params.getEnvironments().split(","))) : new ArrayList<>();
         FileType fileType = params.getFileExtension();
 
-        // make columns present in all exports
-        List<Column> columns = ExperimentFileColumns.getOrderedColumns();
-
         // add columns for requested dataset obsvars and timestamps
         log.debug(logHash + ": fetching experiment for export");
         BrAPITrial experiment = getExperiment(program, experimentId);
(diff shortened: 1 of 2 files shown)


src/main/java/org/breedinginsight/brapi/v2/services/BrAPITrialService.java
```diff
@@ -888,6 +926,10 @@ private List<BrAPIObservation> filterDatasetByEnvironment(
                 .collect(Collectors.toList());
     }
 
+    private boolean isSubEntityDataset(List<BrAPIObservationUnit> ous){
+        return (ous.get(0).getObservationUnitPosition().getObservationLevelRelationships().size() > 2);
```
nickpalladino, line 933: I think the check for subentity would be more robust if we used the datasets array in the trial object additionalinfo. For my test experiment it looks like:
```
"datasets": [
		{
			"id": "eb6387f5-aebe-4aab-a6a1-90b662d963fd",
			"name": "Plot",
			"level": "0"
		},
		{
			"id": "0f2ca27d-c15d-4173-a3b1-e9f9c575fb0e",
			"name": "Plant",
			"level": "1"
		},
		{
			"id": "eb8a41d7-0360-497a-bf1e-3b9be759d211",
			"name": "Berry",
			"level": "1"
		}
	],
``` 
The front end is passing a datasetId in the ExperimentExportQuery query parameter to the end point. It looks like currently it is always the exp unit datasetId but I think the front end could be updated to pass in the datasetId of the selected export dataset. We could then lookup the level for the given datasetId based on the id of the objects in the dataset array, 0 for exp unit and 1 for sub unit.

src/main/java/org/breedinginsight/brapi/v2/services/BrAPITrialService.java
```diff
@@ -810,6 +826,28 @@ private Map<String, Object> createExportRow(
         String observationLvl = ou.getAdditionalInfo().getAsJsonObject().get(BrAPIAdditionalInfoFields.OBSERVATION_LEVEL).getAsString();
         row.put(observationLvl + " " + OBSERVATION_UNIT_ID_SUFFIX, ouId);
 
+        if (isSubEntity) {
+            BrAPIObservationUnitLevelRelationship topLevel = ou.getObservationUnitPosition()
```
nickpalladino: I think you could use the datasets array in the trial additionalinfo here in a similar way as the the isSubEntityDataset recommendation below. I think that would be more robust than relying on the level order in the observationLevelRelationships array.

src/main/java/org/breedinginsight/brapi/v2/services/BrAPITrialService.java
```diff
@@ -188,8 +182,29 @@ public DownloadFile exportObservations(
...
+            columns = ExperimentFileColumns.getOrderedColumnsSubEntity();
+        } else {
+            columns = ExperimentFileColumns.getOrderedColumns();
+        }
+
         //add obsUnitID as dynamic column with observation level appended to header
-        String observationLvl =  ous.get(0).getAdditionalInfo().get(BrAPIAdditionalInfoFields.OBSERVATION_LEVEL).getAsString();
+        if (isSubObs) {
+            //need to add top level obs unit ids as well
+            BrAPIObservationUnitLevelRelationship topLevel = ous.get(0).getObservationUnitPosition()
```
nickpalladino: Looks like getting the top level could be turned into a reusable method since the code is used in another spot below as well.

nickpalladino (changes requested):

src/main/java/org/breedinginsight/brapi/v2/services/BrAPITrialService.java
HMS17: good point, reusable method added

src/main/java/org/breedinginsight/brapi/v2/services/BrAPITrialService.java
nickpalladino: ok keeping this as-is, will revisit brapi server integration issues in future card

nickpalladino (approved):

PR #494, opened 2026-02-03
[BI-2751] - Germplasm entry number weirdness

HMS17:
# Description
**Story:** [BI-2751](https://breedinginsight.atlassian.net/browse/BI-2751) - Germplasm entry number weirdness

Changes to processing germplasm import to match user expectations of behavior.

- Updated error message re parental entry numbers for clarity
- Parental entry numbers will not be saved to additional info on germplasm creation, and not show up when importing the same germplasm in future imports
~- Change in behavior where if a germplasm already has a pedigree in the database, any file pedigree information should be ignored rather than validated and potentially return errors~

NOTE that the import preview not containing entry numbers will only work for germplasm created after this change, a migration would be needed to remove parental entry numbers from additional info of existing germplasm

NOTE germplasm name in the upload file must be correct for error behavior to work properly, making logic solely dependent on accession no will be work of future card

# Dependencies
bi-web:develop

# Testing
Create new program, ensure test cases in linked card now pass

# Checklist:

[BI-2751]: https://breedinginsight.atlassian.net/browse/BI-2751?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/main/java/org/breedinginsight/brapps/importer/services/processors/germplasm/GermplasmProcessor.java
```diff
@@ -405,21 +419,8 @@ private boolean processExistingGermplasm(Germplasm germplasm, ValidationErrors v
             return false;
         }
 
-        // Error conditions:
```
nickpalladino, line 408: Changing this to silently ignore pedigree conflicts instead of rejecting conflicting updates seems like it could be misleading to the user. Was there a reason for removing this?

src/main/java/org/breedinginsight/brapps/importer/services/processors/germplasm/GermplasmProcessor.java
```diff
@@ -793,6 +804,8 @@ else if (germplasmIndexByEntryNo.containsKey(germplasm.getFemaleParentEntryNo())
                 if (commit) {
                     if (femaleParentFound) {
                         brAPIGermplasm.putAdditionalInfoItem(BrAPIAdditionalInfoFields.GERMPLASM_FEMALE_PARENT_GID, femaleParent.getAccessionNumber());
+                        //entry number no longer needed for figuring out parentage, can remove (since same germplasm can have different entry numbers across multiple lists)
+                        brAPIGermplasm.putAdditionalInfoItem(BrAPIAdditionalInfoFields.GERMPLASM_FEMALE_PARENT_ENTRY_NO, null);
```
nickpalladino, line 815: I saw the comment in the PR description about these still showing up without a migration for old records. Is that something we need a card for?

src/main/java/org/breedinginsight/brapps/importer/services/processors/germplasm/GermplasmProcessor.java
HMS17: The intent was that in the case of germplasm that already have a pedigree, pedigree information entered in that row of the file should be ignored and so pedigree conflict checks should not trigger.

src/main/java/org/breedinginsight/brapps/importer/services/processors/germplasm/GermplasmProcessor.java
HMS17: We do still need a card I believe yes, if talking with Shawn probably needed to determine urgency.

src/main/java/org/breedinginsight/brapps/importer/services/processors/germplasm/GermplasmProcessor.java
nickpalladino: I think for normal use cases the pedigree check would not be triggering at undesired times. If a user downloads a germplasm list with pedigree data to do a batch update the pedigree would all be the same and not trigger any errors. If user created a new import file with existing germplasm that had pedigree using GIDs and did not enter any pedigree information that would not trigger any errors either. If a user changed existing pedigree information for an existing germplasm I think they should get an error so they know it is not allowed. Otherwise I think it could be misleading if they were expecting a pedigree update and nothing happened.

src/main/java/org/breedinginsight/brapps/importer/services/processors/germplasm/GermplasmProcessor.java
nickpalladino: Sounds good, can you make a card for that then I'll resolve this comment.

HMS17: https://breedinginsight.atlassian.net/browse/BI-2790

src/main/java/org/breedinginsight/brapps/importer/services/processors/germplasm/GermplasmProcessor.java
HMS17: Agreed to still include validations when pedigree already exists for germplasm, and logic updated.

nickpalladino (approved):
Testing looks good

PR #495, opened 2026-02-13
[BI-2786] Remove code to add treatments to additional info

jloux-brapi:
# Description
**Story:**. [BI-2786](https://breedinginsight.atlassian.net/browse/BI-2786)

Rolled back changes made to support adding treatments to an ObservationUnit's additionalInfo field.  This was originally done to support treatments not being supported in Breedbase.  Since then, DeltaBreed as has switched to BrAPI and grabs treatment data from the brapi server's `observation_unit_treatment` table.

This ended up causing a downstream bug with FB where they were unable to process a JSON object field inside of the additionalInfo of an ObservationUnit.  For now, the FB bug has been fixed by simply not processing json objects, as these additionalInfo attributes are shown to the user in FB and looking at unformatted JSON strings is not desirable.

This change fixes the BI side of the equation, and a migration has also been added in the production server to remove this data out of additionalInfo since it is not relevant anymore.

# Dependencies

brapi-prod-server MR [here](https://github.com/Breeding-Insight/brapi-Java-ProdServer/pull/10).

# Testing
Import an experiment into DeltaBreed with data in the `Treatment` column.

Verify on the brapi DB side that no `"treatments"` attributes are found in the `observation_unit.additional_info` jsonb field for the newly imported observation units.

Verify that the treatments can be seen in the UI for the experiment, and they also show up when said experiment is exported.

# Checklist:

[BI-2786]: https://breedinginsight.atlassian.net/browse/BI-2786?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/main/java/org/breedinginsight/brapi/v2/dao/BrAPIObservationUnitDAO.java
```diff
@@ -432,14 +430,4 @@ private void processObservationUnits(Program program, List<BrAPIObservationUnit>
             }
         }
     }
-
-    private void preprocessObservationUnits(List<BrAPIObservationUnit> brapiObservationUnits) {
-        // add treatments to additional info
-        for (BrAPIObservationUnit obsUnit : brapiObservationUnits) {
-            List<BrAPIObservationTreatment> treatments = obsUnit.getTreatments();
-            if (treatments != null) {
-                obsUnit.putAdditionalInfoItem(BrAPIAdditionalInfoFields.TREATMENTS, treatments);
```
nickpalladino, line 441: I think we should remove TREATMENTS from `BrAPIAdditionalInfoFields`. There are also two other usages that would need to be removed. 

1. `BrAPIObservationUnitDAOTest::testCreateObservationUnitAdditionalInfoSingleTreatmentFactor`

I think for this test we could just instead create the treatment in the observation unit properly and the asserts could check to see that there is no treatments in additionalinfo. We could also skip the additionalinfo check altogether but may make sense to put in because going forward we don't want that.

2. `BrAPIObservationUnitDAO::processObservationUnits`

Looks like we can just remove the section of code there that's checking for the treatments in additionalinfo and copying to the OU treatments.

nickpalladino (changes requested):

src/main/java/org/breedinginsight/brapi/v2/dao/BrAPIObservationUnitDAO.java
jloux-brapi: Done.

nickpalladino (approved):

PR #498, opened 2026-02-24
[BI-2744] DeltaBreed:Create exp, create obs level CI

HMS17:
# Description
**Story:** [BI-2744 - DeltaBreed:Create exp, create obs level ](https://breedinginsight.atlassian.net/browse/BI-2744)

Added logic for experimental import to create observation levels for new exp units.

Added validation to check for more than one exp unit in file and return error if so (previous behavior just took the first row exp unit and silently ignored all other rows).

Note: Testing BI-2791 atop experiments made with these changes revealed a niche case where if an import file was made that forcibly included Exp Units of different casings (ie BaconPancakes and BaconPANCAKES), downloading for a particular environment(s) rather than "All Environments" resulted in the download failing due to a null pointer exception because in BrAPITrialService.java:exportObservations, the hash "entry" would get an additional value with a null key for the alternate casing version of the Exp Unit. Since this logic is dependent upon the Exp Unit value stored in additionalInfo, it was determined that the fix should be covered in BI-2778, which will deprecate use of additionalInfo for level, and that case tested there.

# Dependencies
bi-web: [feature/BI-2744](https://github.com/Breeding-Insight/bi-web/pull/458)

# Testing
- Create experiment file with multiple exp unit values
-- Attempt to upload file
-- Error message should appear and file upload should be prevented

- Create experiment file with new exp unit value (not Plot)
-- Upload file
-- Ensure Experiment Preview displays the exp unit value as capitalized regardless of what case the value was in file
-- Confirm Import and ensure file imports successfully
-- Ensure Experiment Details displays the exp unit value as capitalized regardless of what case the value was in file
-- Check that the GET /observationlevels call for that program returns the new exp unit value

- Create experiment file with same exp unit value as the first upload
-- Upload file
-- Confirm Import and ensure file imports successfully

# Checklist:

### DIFF
--- a/src/main/java/org/breedinginsight/brapps/importer/services/processors/experiment/create/workflow/steps/CommitPendingImportObjectsStep.java
+++ b/src/main/java/org/breedinginsight/brapps/importer/services/processors/experiment/create/workflow/steps/CommitPendingImportObjectsStep.java
@@ -16,6 +16,7 @@
  */
 package org.breedinginsight.brapps.importer.services.processors.experiment.create.workflow.steps;
 
+import io.micronaut.http.HttpResponse;
 import io.micronaut.http.server.exceptions.InternalServerException;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.collections4.map.CaseInsensitiveMap;
@@ -42,6 +43,7 @@ import org.breedinginsight.brapps.importer.services.processors.ProcessorData;
 import org.breedinginsight.brapps.importer.services.processors.experiment.ExperimentUtilities;
 import org.breedinginsight.brapps.importer.services.processors.experiment.create.model.PendingData;
 import org.breedinginsight.brapps.importer.services.processors.experiment.create.model.ProcessContext;
+import org.breedinginsight.model.DatasetLevel;
 import org.breedinginsight.model.Program;
 import org.breedinginsight.model.ProgramLocation;
 import org.breedinginsight.model.Trait;
@@ -72,6 +74,7 @@ public class CommitPendingImportObjectsStep {
     private final BrAPIObservationUnitDAO brAPIObservationUnitDAO;
     private final ProgramLocationService locationService;
     private final OntologyService ontologyService;
+    private final BrAPIObservationLevelDAO brAPIObservationLevelDAO;
 
     @Inject
     public CommitPendingImportObjectsStep(BrAPIListDAO brAPIListDAO,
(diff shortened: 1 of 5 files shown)


jloux-brapi:
Tested with FB.  Endless fetching error has been resolved with this change!

src/main/java/org/breedinginsight/brapps/importer/services/processors/experiment/create/workflow/steps/CommitPendingImportObjectsStep.java
```diff
@@ -243,6 +251,22 @@ public void process(ProcessContext processContext, ProcessedData processedData)
...
     }
 
+    //Check if the experimental unit associated with the trial does not exist in the system and if so create a new observation level
+    private void createObservationLevel(String trialName, Map<String, String> expUnitByTrialName, Program program) throws ApiException, InternalServerException {
+        String expUnit = expUnitByTrialName.get(trialName).toLowerCase();
+        String programDbId = program.getBrapiProgram() != null ? program.getBrapiProgram().getProgramDbId() : null;
+        HttpResponse<String> levelResponse = brAPIObservationLevelDAO.createObservationLevelName(program, expUnit, DatasetLevel.EXP_UNIT, programDbId);
+
+        if (levelResponse.getStatus().getCode() == 409) {
+            log.info(String.format("Level %s already exists in database", expUnit));
```
jloux-brapi: There's an incoming ticket to make each level name unique against its name, order, and program id.  This will be necessary to preserve the order of whatever level name is the top-level or sub-level entity for multiple experiments in a program.  This means that this could error out based off of not just what the level name is.

I think in these logs (and elsewhere) we should also log the order and program id for clarity on what went wrong, and what the request might have looked like.  In this case I guess it's pretty obvious that the level order will always be 0, but I still think it would be helpful from a log skimming perspective.

jloux-brapi:
```suggestion
            log.info(String.format("Level with name=%s, order=%s, programDbId=%s already exists in database", expUnit, DatasetLevel.EXP_UNIT, programDbId));
```

src/main/java/org/breedinginsight/brapps/importer/services/processors/experiment/create/workflow/steps/CommitPendingImportObjectsStep.java
```diff
@@ -243,6 +251,22 @@ public void process(ProcessContext processContext, ProcessedData processedData)
...
+        } else if (levelResponse.getStatus().getCode() == 200) {
+            log.info(String.format("Level %s created in database", expUnit));
```
jloux-brapi:
```suggestion
            log.info(String.format("Level with name=%s, order=%s, programDbId=%s created in database", expUnit, DatasetLevel.EXP_UNIT, programDbId));
```

jloux-brapi (changes requested):
Just want some extra info on the level names being logged, suggested changes included.

HMS17:
> Just want some extra info on the level names being logged, suggested changes included.

Good suggestions, committed!

jloux-brapi (approved):

PR #499, opened 2026-02-26
[BI-2791] Exp File Name Improvement

HMS17:
# Description
**Story:** [BI-2791 - Exp File Name Improvement](https://breedinginsight.atlassian.net/browse/BI-2791)

Improvements to file name of downloaded experiments:
- Abbreviating “All_Environments” to “All_Env
- Naming the file with the associated level name instead of “Observation_Dataset”

# Dependencies
bi-web: develop

# Testing
Download:
Top level dataset, all environments
Top level dataset, specific environment
Sub entity dataset, all environments
Sub entity dataset, specific environment

File names for individual files with all environments should be in the format Experiment Name_Dataset Name_All_Env_Timestamp
File names for individual files for a specific environment should be in the format Experiment Name_Dataset Name_Environment Name_Timestamp
File names for zip files should be in format Experiment Name_Dataset Name_Timestamp

# Checklist:

- [x] I have either updated the source of truth or arranged for update with product owner if needed https://breedinginsight.atlassian.net/wiki/spaces/BI/pages/1559953409/Source+of+Truth

### DIFF
--- a/src/main/java/org/breedinginsight/brapi/v2/services/BrAPITrialService.java
+++ b/src/main/java/org/breedinginsight/brapi/v2/services/BrAPITrialService.java
@@ -292,8 +292,7 @@ public class BrAPITrialService {
                 List<Map<String, Object>> rows = entry.getValue();
                 sortDefaultForExportRows(rows);
                 StreamedFile streamedFile = FileUtil.writeToStreamedFile(columns, rows, fileType, SHEET_NAME);
-                // TODO: [BI-2183] remove hardcoded datasetName, use observation level.
-                String name = makeFileName(experiment, program, studyByDbId.get(entry.getKey()).getStudyName(), "Observation Dataset") + fileType.getExtension();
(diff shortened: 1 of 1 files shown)


jloux-brapi (approved):
Tested, works as described.

I don't think this is part of the ticket but I'm wondering if there's any thought into the process of how it should be named when there are multiple envs selected rather than just one.  Right now no env name shows up when that happens, just seems like it has the latest year or something.  Could be worth asking Shawn what she thinks of that.

HMS17:
> Tested, works as described.
> 
> I don't think this is part of the ticket but I'm wondering if there's any thought into the process of how it should be named when there are multiple envs selected rather than just one. Right now no env name shows up when that happens, just seems like it has the latest year or something. Could be worth asking Shawn what she thinks of that.

It is a good question, I think though that since the env names are in the files within the greater zip file, that would be okay? Especially since we don't want to make the zip file name excessively long.

PR #500, opened 2026-03-06
[BI-2778] Deprecate additionalinfo usage for level and use obs unit lvl

HMS17:
# Description
**Story:** [BI-2778 - Deprecate additionalinfo usage for level and use obs unit level](https://breedinginsight.atlassian.net/browse/BI-2778)

Removed calls to set observation level info in additionalInfo and replaced calls to get observation level info from additionalInfo with calls to the observation level associated with the observation object.

This should not be merged until after the migration of BI-2779 is merged.

# Dependencies
bi-web: [feature/BI-2778](https://github.com/Breeding-Insight/bi-web/pull/459) (old)
bi-web: [feature/BI-2778](https://github.com/Breeding-Insight/bi-web/pull/460) (new)

# Testing
Test areas related to experiments to ensure nothing is broken:

-- Upload experiment
-- Create sub entity dataset
-- View experiment
-- Download top level experiment dataset
-- Download sub entity experiment dataset
-- Download case mentioned in comments of card

# Checklist:


src/main/java/org/breedinginsight/brapi/v2/services/BrAPITrialService.java
```diff
@@ -995,7 +989,7 @@ private void sortDefaultForObservationUnit(List<BrAPIObservationUnit> ous) {
 
         if (isSubEntityDataset(ous)) {
             Comparator<BrAPIObservationUnit> subUnitComparator = Comparator.comparing(BrAPIObservationUnit::getObservationUnitName, new IntOrderComparator());
-            Comparator<BrAPIObservationUnit> ouNameComparator = Comparator.comparing(row -> (row.getAdditionalInfo().get(BrAPIAdditionalInfoFields.EXP_UNIT_ID).toString()), new IntOrderComparator());
+            Comparator<BrAPIObservationUnit> ouNameComparator = Comparator.comparing(row -> (getTopLevel(row).getLevelName()), new IntOrderComparator());
```
jloux-brapi: Is it ok to swap the logic for exp unit id with level name?

src/main/java/org/breedinginsight/brapi/v2/services/BrAPITrialService.java
```diff
@@ -897,13 +891,13 @@ private Map<String, Object> createExportRow(
                 String topLvlOuId = Utilities.removeProgramKeyAndUnknownAdditionalData(topLevel.getLevelCode(), program.getKey());
                 row.put(topLvlName + " " + OBSERVATION_UNIT_ID_SUFFIX, topLvlOuId);
             }
-            row.put(ExperimentObservation.Columns.EXP_UNIT_ID, ou.getAdditionalInfo().get(BrAPIAdditionalInfoFields.EXP_UNIT_ID).getAsString());
+            row.put(ExperimentObservation.Columns.EXP_UNIT_ID, StringUtils.capitalize(getTopLevel(ou).getLevelName()));
```
jloux-brapi: Is it ok to swap the logic for exp unit id with level name?

src/main/java/org/breedinginsight/brapi/v2/services/BrAPITrialService.java
HMS17: I do not think so, EXP_UNIT and EXP_UNIT_ID are two separate values. For the sort, we want to know the top level dataset observation unit associated with the sub entity unit, not just the level name.

src/main/java/org/breedinginsight/brapi/v2/services/BrAPITrialService.java
HMS17: For EXP_UNIT column in a row we want level name. EXP_UNIT_ID column meanwhile is the id of the specific observation unit from the top level dataset.

src/main/java/org/breedinginsight/brapi/v2/services/BrAPITrialService.java
jloux-brapi: Then I'm confused, why is this comparing level name then instead of exp_unit_id.  And if it is comparing level name, why is it using an IntOrderComparator?

src/main/java/org/breedinginsight/brapi/v2/services/BrAPITrialService.java
jloux-brapi: RIght, but this new commit is putting the level name in the exp unit id column is it not?

HMS17: Ah I misinterpreted, yeah good catch this is incorrect.

src/main/java/org/breedinginsight/brapi/v2/services/BrAPITrialService.java
HMS17: Ah, I don't know why for some reason I interpreted this as you suggesting a change rather than pointing out my change was invalid. Good catch, I'll fix that.

jloux-brapi (approved):

jloux-brapi (changes requested):
Definitely some things going wrong with the testing, not quite sure what.

The changes I suggested seemed to introduce an NPE, when I reverted that, I am still noticing some issues with the UI I have never seen before.

For example, the sub entity data Obs Unit Id column says `Undefined`  Obs Unit Id where it should say the level name.

jloux-brapi (approved):
Looks and tested good, checked the BE, don't see any obs lvls in ou additional info.

PR #506, opened 2026-04-01
[BI-2814] Create Sub Entity Dataset: Update OUs to supply levelNameDbIds

jloux-brapi:
# Description
**Story:** [BI-2814](https://breedinginsight.atlassian.net/browse/BI-2814)

- Consolidated relevant code into DatasetService for use in creation experiment use case
- Updated to BrAPI Java Client 2.2
- Added methods to get data from the updated client in BrAPIObservationLevelDAO

# Dependencies

[BI-2826 MR](https://github.com/Breeding-Insight/brapi/pull/227)

# Testing
* Create a program then do the following: 
* * Create a new experiment with any level name
* * Create a sub-entity dataset with any level name (except top level name of this experiment, this will always result in an error as intended)
* * Verify no errors occur on creating the sub entity dataset, and that the data can be viewed for the sub entity dataset in the experiment, and the sub entity dataset data can also be exported without issue

* In a program where there already exists an experiment with both a top level dataset and a sub entity dataset: 
* * Create a new experiment with a different top level name (except that of the sub entity dataset of an existing experiment in the program, that is still broken and fixed will be fixed via https://breedinginsight.atlassian.net/browse/BI-2813)
* * Create a sub-entity dataset that has the same name as an existing top level dataset in a different experiment in the program.
* * Verify no errors occur on creating the sub entity dataset, and that the data can be viewed for the sub entity dataset in the experiment, and the sub entity dataset data can also be exported without issue

* In a program where there already exists an experiment with both a top level dataset and a sub entity dataset: 
* * Create a new experiment with a different top level name (except that of the sub entity dataset of an existing experiment in the program, that is still broken and fixed will be fixed via https://breedinginsight.atlassian.net/browse/BI-2813)
* * Create a sub-entity dataset that has the same name as an existing top level dataset in a different experiment in the program.
* * Verify no errors occur on creating the sub entity dataset, and that the data can be viewed for the sub entity dataset in the experiment, and the sub entity dataset data can also be exported without issue

# Checklist:

- [X] I have performed a self-review of my own code
- [X] I have tested my code and ensured it meets the acceptance criteria of the story
- [X] I have create/modified unit and/or integration tests to cover this change or tests are not applicable
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have either updated the source of truth or arranged for update with product owner if needed https://breedinginsight.atlassian.net/wiki/spaces/BI/pages/1559953409/Source+of+Truth

[...]

PR #507, opened 2026-04-02
[BI-2813] Set levelNameDbIds in OUs in experiment creation use case

jloux-brapi:
# Description
**Story:** [BI-2813](https://breedinginsight.atlassian.net/browse/BI-2813)

* Updated observation level creation logic in the create experiment workflow to work similarly and use similar methods as sub entity workflow
* Added some comments to CommitPendingImportObjectsStep regarding how it could be improved to reflect the actual use case it runs, and some deterministic logic which is needed to support the observation level creation logic supporting that use case.

# Dependencies
[BI-2814](https://github.com/Breeding-Insight/bi-api/pull/506) should be merged first, and then this ticket/MR should be pointed to `develop`
# Testing
Create a program then do the following: 

* Create a new experiment with any level name
* Create a sub-entity dataset with any level name (except top level name of this experiment, this will always result in an error as intended)
* Verify no errors occur on creating the datasets, and that the data can be viewed for the datasets in the experiment, and the dataset data can also be exported without issue

In a program where there already exists an experiment with both a top level dataset and a sub entity dataset: 

* Create a new experiment with top level name that has the same name as the existing sub entity dataset name created in the first experiment
* Create a sub-entity dataset that has the same name as an existing top level dataset in a different experiment in the program.
* Verify no errors occur on creating the datasets, and that the data can be viewed for the datasets in the experiment, and the dataset data can also be exported without issue

In a program where there already exists an experiment with both a top level dataset and a sub entity dataset: 
* Create a new experiment with top level name that has the same name as the existing sub entity dataset name created in the first experiment
* Create a sub-entity dataset with a name that isn’t shared by any of the other existing sub entity datasets for the experiments in this program.
* Verify no errors occur on creating the datasets, and that the data can be viewed for the datasets in the experiment, and the dataset data can also be exported without issue

# Checklist:

[BI-2813]: https://breedinginsight.atlassian.net/browse/BI-2813?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[BI-2814]: https://breedinginsight.atlassian.net/browse/BI-2814?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/org/breedinginsight/brapi/v2/services/BrAPIObservationLevelService.java
+++ b/src/main/java/org/breedinginsight/brapi/v2/services/BrAPIObservationLevelService.java
@@ -1,49 +1,28 @@
 package org.breedinginsight.brapi.v2.services;
 
 import lombok.extern.slf4j.Slf4j;
-import org.apache.commons.lang3.tuple.ImmutablePair;
-import org.apache.commons.lang3.tuple.Pair;
-import org.brapi.client.v2.ApiResponse;
 import org.brapi.client.v2.model.exceptions.ApiException;
-import org.brapi.client.v2.modules.phenotype.ObservationUnitsApi;
 import org.brapi.v2.model.pheno.BrAPIObservationUnitHierarchyLevel;
-import org.brapi.v2.model.pheno.BrAPIObservationUnitLevelRelationship;
-import org.brapi.v2.model.pheno.response.BrAPIObservationLevelListResponse;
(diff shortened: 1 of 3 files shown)


nickpalladino (approved):
Testing looked good

PR #513, opened 2026-04-27
BI-2789 - Create Germplasm: loosen breeding method and source requirement

humsika:
# Description
Link to the story in Jira: https://breedinginsight.atlassian.net/browse/BI-2789?atlOrigin=eyJpIjoiZjEzZmIyMjNmNWVmNDIyYjk5ZjIxMzM4MTJiZDY3ZTkiLCJwIjoiaiJ9

This PR updates germplasm import validation behavior for BI-2789.

The change removes the unconditional required validation for Source and makes Breeding Method conditionally required only when parent information is provided through parent GIDs or parent entry numbers.

This PR also updates existing germplasm append behavior so that Source and Breeding Method can be appended when the corresponding database field is blank, while ensuring existing populated values are not overwritten.

In addition, backend tests were added and updated to cover the new validation and append scenarios introduced by this story.

# Dependencies
bi-web: develop
bi-api: feature/BI-2789

No separate frontend changes are included in this PR.

# Testing
- Added/updated germplasm import backend tests for BI-2789 scenarios.
- Verified locally that the added/updated tests are passing.
- Tested new germplasm import with no parent data to confirm blank Breeding Method is allowed.
- Tested germplasm import with parent GIDs / parent entry numbers to confirm Breeding Method is required when parent data is present.
- Tested blank Source handling to confirm it is no longer always required.
- Tested existing germplasm append behavior to confirm blank Source / Breeding Method fields can be filled.
- Tested existing germplasm behavior to confirm populated Source / Breeding Method values are not overwritten.

# Checklist:

- [x] I have either updated the source of truth or arranged for update with product owner if needed https://breedinginsight.atlassian.net/wiki/spaces/BI/pages/1559953409/Source+of+Truth

### DIFF
--- a/src/main/java/org/breedinginsight/brapps/importer/model/base/Germplasm.java
+++ b/src/main/java/org/breedinginsight/brapps/importer/model/base/Germplasm.java
@@ -57,12 +57,10 @@ public class Germplasm implements BrAPIObject {
     private String germplasmName;
 
     @ImportFieldType(type= ImportFieldTypeEnum.TEXT)
-    @ImportMappingRequired
     @ImportFieldMetadata(id="breedingMethod", name="Breeding Method", description = "The breeding method name or code")
     private String breedingMethod;
 
     @ImportFieldType(type= ImportFieldTypeEnum.TEXT)
-    @ImportMappingRequired
     @ImportFieldMetadata(id="germplasmSource", name="Source", description = "The germplasm origin. If External UID present, assumed to be the source associated with the External UID.")
     private String germplasmSource;
 
@@ -165,7 +163,8 @@ public class Germplasm implements BrAPIObject {
      * @param updatePedigree flag indicating if pedigree should be updated
      * @return mutated indicator
      */
-    public boolean updateBrAPIGermplasm(BrAPIGermplasm germplasm, Program program, UUID listId, boolean commit, boolean updatePedigree) {
+     public boolean updateBrAPIGermplasm(BrAPIGermplasm germplasm, Program program, UUID listId, boolean commit,
+                                         boolean updatePedigree, ProgramBreedingMethodEntity breedingMethod) {
 
         boolean mutated = false;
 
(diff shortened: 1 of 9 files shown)


davedrp:
Developer test passed

src/main/java/org/breedinginsight/brapps/importer/model/base/Germplasm.java
```diff
@@ -165,7 +163,8 @@ public static String constructGermplasmListName(String listName, Program program
      * @param updatePedigree flag indicating if pedigree should be updated
      * @return mutated indicator
      */
-    public boolean updateBrAPIGermplasm(BrAPIGermplasm germplasm, Program program, UUID listId, boolean commit, boolean updatePedigree) {
+     public boolean updateBrAPIGermplasm(BrAPIGermplasm germplasm, Program program, UUID listId, boolean commit,
```
davedrp: `listId`  is never used.

src/main/java/org/breedinginsight/brapps/importer/services/processors/germplasm/GermplasmProcessor.java
```diff
@@ -397,7 +384,8 @@ private Germplasm removeBreedingMethodBlanks(Germplasm germplasm) {
         return germplasm;
     }
 
-    private boolean processExistingGermplasm(Germplasm germplasm, ValidationErrors validationErrors, List<BrAPIImport> importRows, Program program, UUID importListId, boolean commit, PendingImport mappedImportRow, int rowIndex) {
+    private boolean processExistingGermplasm(Germplasm germplasm, ValidationErrors validationErrors, List<BrAPIImport> importRows, Map<String, ProgramBreedingMethodEntity> breedingMethods,
```
davedrp: The return value is never used.  
- Should return `void`
- Each `return` value would also need to be changed.

davedrp (changes requested):
There are a couple of pre-existing issues.  While they are pre-existing, they where in code that was newly updated.  So, I think, it is a good idea to take this opportunity to address the issues.

src/main/java/org/breedinginsight/brapps/importer/model/base/Germplasm.java
humsika: updated the code as per suggestion, thanks

src/main/java/org/breedinginsight/brapps/importer/services/processors/germplasm/GermplasmProcessor.java
humsika: updated the code as per suggestion, thanks

davedrp (approved):

PR #520, opened 2026-05-28
BI-2887: Germplasm download error with blank source or breeding method

humsika:
# Description
**Story:** [BI-2887](https://breedinginsight.atlassian.net/browse/BI-2887?atlOrigin=eyJpIjoiM2M3MjUzZGQ4YTc2NDFiYmIyNTU4NGYzYzU1ODcxNGUiLCJwIjoiaiJ9)

This PR fixes a backend-only regression in germplasm export/download behavior introduced after BI-2789 made `Source` and `Breeding Method` optional for records without parent information.

Previously, germplasm downloads could fail when a record had:
- a blank `Breeding Method`
- a blank `Source`
- both fields blank

This change makes the export processing null-safe so downloads succeed and blank optional values are exported as blank cells.

# Dependencies
bi-api: bug/BI-2887
bi-web: develop

# Testing
- Verified download succeeds in all cases
- Verified blank optional fields are exported as blank cells
- Verified export column order is unchanged
- Verified germplasm-list export preserves list order and entry numbers

# Checklist:

- [ ] I have either updated the source of truth or arranged for update with product owner if needed https://breedinginsight.atlassian.net/wiki/spaces/BI/pages/1559953409/Source+of+Truth

[BI-2887]: https://breedinginsight.atlassian.net/browse/BI-2887?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/test/java/org/breedinginsight/services/BrAPIGermplasmServiceUnitTest.java
+++ b/src/test/java/org/breedinginsight/services/BrAPIGermplasmServiceUnitTest.java
@@ -185,12 +185,11 @@ public class BrAPIGermplasmServiceUnitTest extends DatabaseTest {
         testGermplasm.setAccessionNumber("1");
 
         JsonObject additionalInfo = new JsonObject();
-        additionalInfo.addProperty(GERMPLASM_BREEDING_METHOD, false);
         additionalInfo.addProperty(MALE_PARENT_UNKNOWN, false);
         additionalInfo.addProperty(FEMALE_PARENT_UNKNOWN, false);
 
         testGermplasm.setAdditionalInfo(additionalInfo);
-        testGermplasm.setExternalReferences(null);
+        testGermplasm.setExternalReferences(new ArrayList<>());
 
         germplasmService = new BrAPIGermplasmService(listDAO, programService, germplasmDAO);
 
@@ -203,7 +202,7 @@ public class BrAPIGermplasmServiceUnitTest extends DatabaseTest {
         assertEquals(1, processedData.size());
         assertEquals("Germplasm A", processedData.get(0).get("Germplasm Name"));
         assertEquals(1, processedData.get(0).get("Entry No"));
-        assertTrue(processedData.get(0).containsKey("Breeding Method"));
+        assertFalse(processedData.get(0).containsKey("Breeding Method"));
         assertFalse(processedData.get(0).containsKey("Source"));
         assertFalse(processedData.get(0).containsKey("External UID"));
     }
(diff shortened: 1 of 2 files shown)


src/test/java/org/breedinginsight/services/BrAPIGermplasmServiceUnitTest.java
```diff
@@ -165,12 +166,94 @@ public void getGermplasmListExport() {
...
+    public void processListDataAllowsBlankOptionalFieldsForProgramExport() { // NEW: covers full germplasm export path
+        Program testProgram = new Program();
+        testProgram.setKey("TEST");
+
+        BrAPIGermplasm testGermplasm = new BrAPIGermplasm();
+        testGermplasm.setGermplasmName("Germplasm A");
+        testGermplasm.setAccessionNumber("1");
+
+        JsonObject additionalInfo = new JsonObject();
+        additionalInfo.addProperty(GERMPLASM_BREEDING_METHOD, false);
```
nickpalladino: For clarity I think this should be set to a breeding method like "Autopolyploid"

nickpalladino (changes requested):
Looks good, just had a request on value set in test for clarity. Also updated the acceptance criteria slightly in the card so we don't have to deal with the External UID case when having a blank source and will do that in a future card.

src/test/java/org/breedinginsight/services/BrAPIGermplasmServiceUnitTest.java
humsika: I intentionally excluded the Breeding Method field here because this test is meant to cover the case where that field is not set in the data being downloaded.

nickpalladino: It looks like this would be adding the breeding method property and setting the value to false. There is an `assertTrue(processedData.get(0).containsKey("Breeding Method"));` at the end.

humsika: Thanks for the clarification. That makes sense, I’ll make the changes accordingly.

humsika: Made changes as per the instructions.

nickpalladino (approved):

PR #521, opened 2026-05-29
Apply BrAPI-server side Filtering/Sorting/Pagination to BI Experiments

jloux-brapi:
# Description
**Story:** [BI-2861](https://breedinginsight.atlassian.net/browse/BI-2861)

To support the removal of the cache for TrialEntities, a generalized implementation of filtering and sorting was created for the prod server. 

This body of changes modifies bi-api to utilize these new changes via search requests with the newly implemented SortBy and FilterBy request models that are now accessible via an updated brapi-java-client/model.

The changes to bi-api only affect Experiments/Trials, and should namely only really affect the data table seen in a Program's Experiments tab.

Included updates:

- Upgrade to new brapi-java-client version
- Pass ExperimentQuery parameters further into bi-api business logic so they can be handled by the brapi server
- ExperimentQuery parameters now used to build BrAPISearchRequest which is used to call BrAPI Trials Search POST from java client.
- Created generalized code for usage in other search requests in BrAPIDAOUtil
- Created BrAPISortFilterMapper abstract class to be implemented by entities that need to be mapped.  These implementations will define the mapping between the bi-web request names to the brapi-prod-server request names for the columns that need to be searched on
- Updated unit tests to utilize the program search for experiments instead of the existing implementation which is used by the fe to filter.  This will separate these concerns

# Dependencies
- [brapi-java-client and model](https://github.com/Breeding-Insight/brapi/pull/229)
- [java-prod-server](https://github.com/Breeding-Insight/brapi-Java-ProdServer/pull/17)
# Testing
- Test filtering and sorting of Trials by messing around with the different functionalities of the data table in the `Experiments` tab of a Program.  Mess with filtering, testing case insensitivity, and different data types.  For sorting, test the order of different columns and ASC vs DESC order.  Mess with paging, try selecting different pages, and different number of records per page

# Checklist:

[BI-2861]: https://breedinginsight.atlassian.net/browse/BI-2861?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/org/breedinginsight/brapi/v2/BrAPITrialsController.java
+++ b/src/main/java/org/breedinginsight/brapi/v2/BrAPITrialsController.java
@@ -62,8 +62,6 @@ public class BrAPITrialsController {
 
     private final String referenceSource;
     private final BrAPITrialService experimentService;
-    // TODO: See if it's feasible to remove this mapper entirely
-    private final ExperimentQueryMapper experimentQueryMapper;
     private final SecurityService securityService;
     private final ProgramUserService programUserService;
     private final ExperimentalCollaboratorService experimentalCollaboratorService;
@@ -71,14 +69,12 @@ public class BrAPITrialsController {
 
     @Inject
     public BrAPITrialsController(BrAPITrialService experimentService,
-                                 ExperimentQueryMapper experimentQueryMapper,
                                  @Property(name = "brapi.server.reference-source") String referenceSource,
                                  SecurityService securityService,
                                  ProgramUserService programUserService,
                                  ExperimentalCollaboratorService experimentalCollaboratorService,
                                  ProgramService programService) {
         this.experimentService = experimentService;
-        this.experimentQueryMapper = experimentQueryMapper;
         this.referenceSource = referenceSource;
         this.securityService = securityService;
         this.programUserService = programUserService;
@@ -103,7 +99,6 @@ public class BrAPITrialsController {
             // If the program user is an experimental collaborator, filter results for only authorized experiments.
             Optional<ProgramUser> experimentalCollaborator = programUserService.getIfExperimentalCollaborator(programId, securityService.getUser().getId());
             if (experimentalCollaborator.isPresent()) {
-                // TODO: Modify this code to use brapi side filter, pagination and sort
                 List<UUID> authorizedExperimentIds = experimentalCollaboratorService.getAuthorizedExperimentIds(experimentalCollaborator.get().getId());
 
                 BrAPITrialListResponse brapiResponse = experimentService.searchTrials(program.get(), authorizedExperimentIds, queryParams);
--- a/src/main/java/org/breedinginsight/brapi/v2/BrAPITrialsController.java
+++ b/src/main/java/org/breedinginsight/brapi/v2/BrAPITrialsController.java
@@ -161,7 +161,7 @@ public class BrAPITrialsController {
         return HttpResponse.notFound();
     }
 
-    // TODO: Remove for trialDbId once cache is removed for that entity
+    // TODO: Verify that the programDbId does not need to be reset and that the BrAPI programDbId can be used instead [BI-2931]
     private void setDbIds(BrAPITrial trial) {
         trial.programDbId(Utilities.getExternalReference(trial.getExternalReferences(), Utilities.generateReferenceSource(referenceSource, ExternalReferenceSource.PROGRAMS))
                                  .orElseThrow(() -> new IllegalStateException("No BI external reference found"))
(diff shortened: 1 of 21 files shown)


HMS17 (approved):
Code looks good, and light testing on local with small datasets passed

src/main/java/org/breedinginsight/brapi/v2/BrAPITrialsController.java
```diff
@@ -105,13 +99,21 @@ public HttpResponse<Response<DataResponse<List<BrAPITrial>>>> getExperiments(
             // If the program user is an experimental collaborator, filter results for only authorized experiments.
             Optional<ProgramUser> experimentalCollaborator = programUserService.getIfExperimentalCollaborator(programId, securityService.getUser().getId());
             if (experimentalCollaborator.isPresent()) {
-                List<UUID> experimentIds = experimentalCollaboratorService.getAuthorizedExperimentIds(experimentalCollaborator.get().getId());
-                List<BrAPITrial> authorizedExperiments = experimentService.getTrialsByExperimentIds(program.get(), experimentIds).stream().peek(this::setDbIds).collect(Collectors.toList());
-                return ResponseUtils.getBrapiQueryResponse(authorizedExperiments, experimentQueryMapper, queryParams, searchRequest);
+                List<UUID> authorizedExperimentIds = experimentalCollaboratorService.getAuthorizedExperimentIds(experimentalCollaborator.get().getId());
+
+                BrAPITrialListResponse brapiResponse = experimentService.searchTrials(program.get(), authorizedExperimentIds, queryParams);
```
nickpalladino: If experimental collaborator is user in program but not a collaborator in any experiments, think this could expose all program trials because search request won't have any trialDbIds set, should just return empty result in that case.

humsika (approved):
Code looks good to me.

nickpalladino:
@codex review

src/main/java/org/breedinginsight/utilities/response/ResponseUtils.java
```diff
@@ -129,6 +131,19 @@ private static <T> HttpResponse<Response<DataResponse<T>>> processBrapiResponse(
         return HttpResponse.ok(new Response(metadata, new DataResponse(paginationResult.getLeft())));
     }
 
+    public static <T> HttpResponse<Response<DataResponse<T>>> getBrapiQueryResponse(List<T> data,
+                                                                                    BrAPIResponse brAPIResponse,
+                                                                                    BrapiQuery queryParams) {
+        Pagination pagination = new Pagination();
+        pagination.setTotalCount(brAPIResponse.getMetadata().getPagination().getTotalCount());
+        pagination.setCurrentPage(queryParams.getPage());
+        pagination.setPageSize(queryParams.getPageSize());
```
nickpalladino: totalPages isn't set in pagination

nickpalladino: I think currentPage and pageSize also be set from the brAPIResponse, seems like it's possible queryParams may not be set.

src/main/java/org/breedinginsight/brapi/v2/dao/impl/BrAPITrialDAOImpl.java
```diff
@@ -247,4 +267,57 @@ public void deleteBrAPITrial(Program program, BrAPITrial trial, boolean hard) th
...
+
+        if (!brAPIFilterBy.isEmpty()) {
+            searchRequest.setFilterBy(brAPIFilterBy);
+        }
+
+        // Set SortBy
+        List<BrAPISortBy> brAPISortBy = new ArrayList<>();
+
+        if (StringUtils.isNotBlank(experimentQuery.getSortField()) && brAPITrialsMapper.exists(experimentQuery.getSortField())) {
+            brAPISortBy.add(brAPIDAOUtil.constructSortBy(brAPITrialsMapper.getBrAPIName(experimentQuery.getSortField()), experimentQuery.getSortOrder().toString()));
```
nickpalladino, line 311: If getSortOrder() is null the toString() will result in a NullPointerException. Think if sort order is null we should default to ASC

nickpalladino (changes requested):

src/main/java/org/breedinginsight/brapi/v2/dao/impl/BrAPITrialDAOImpl.java
jloux-brapi: I fixed this by setting `sortOrder = SortOrder.ASC` in parent class of `ExperimentQuery`, `BrapiQuery`.

It should not be able to be null now unless a caller requests it set to null.

src/main/java/org/breedinginsight/utilities/response/ResponseUtils.java
jloux-brapi: Fixed, and everything seems to be working properly

jloux-brapi: Fixed

src/main/java/org/breedinginsight/brapi/v2/BrAPITrialsController.java
jloux-brapi: Addressed

nickpalladino (approved):

PR #523, opened 2026-06-08
BI-2848 - New Genotyping Page Backend: linkage to sample submission

humsika:
# Description
**Story:** [BI-2848](https://breedinginsight.atlassian.net/browse/BI-2848?atlOrigin=eyJpIjoiZTIzZWMxNzQxNjFlNDljNmIzY2EzM2NiZGE2ZDhjODMiLCJwIjoiaiJ9)

This change adds backend support for the new Genotyping page by linking successful genotype imports to sample submissions and exposing a new endpoint for the frontend.

Summary of changes made:
- Added support for linking `sample_submission` and `importer_import` through the new `genotype_import` join table
- Added GET endpoint: `/programs/{programId}/geno/imports`
- Added response fields needed by frontend, including:
  - `sampleSubmissionId`
  - `projectNameForSampleSubmission`
  - `sampleSubmissionCreatedBy`
  - `genotypingFileName`
  - `genotypingImportDate`
  - `genotypingImportBy`
- Added paging, sorting, and filtering support for the endpoint
- Updated the genotype import flow so only **successful** genotype imports are linked/shown on the Genotyping page
- Updated/added controller and service tests for the new endpoint and service behavior

# Dependencies
- DB migration for `genotype_import` must be applied
- jOOQ generated classes should be up to date
- bi-web (frontend): feature/BI-2679
- bi-api (backend): feature/BI-2848

# Testing
Validate the Backend/API by performing the following through manual testing:

1. Run DB migration and ensure generated jOOQ classes are updated
2. Upload a valid genotype VCF for a sample submission
3. Verify the record appears in:
   - `genotype_import`
   - `GET /v1/programs/{programId}/geno/imports`
4. Verify endpoint supports:
   - paging
   - sorting
   - filtering
5. Verify response includes `sampleSubmissionId` for frontend linking
6. Verify failed/invalid genotype imports do not appear in the Genotyping page response

Example API checks:
- Base:
  - `GET /v1/programs/{programId}/geno/imports?page=1&pageSize=10`
- Sorted:
  - `GET /v1/programs/{programId}/geno/imports?page=1&pageSize=10&sortField=genotypingImportDate&sortOrder=DESC`
- Filtered:
  - `GET /v1/programs/{programId}/geno/imports?page=1&pageSize=10&projectNameForSampleSubmission=Sample1`
  - `GET /v1/programs/{programId}/geno/imports?page=1&pageSize=10&genotypingFileName=vcf`

Expected behavior:
- Successful imports are returned
- Failed imports are not returned
- Invalid program returns `404`
- Unauthorized/forbidden access returns appropriate auth response
- Empty result set returns standard empty `data` response

# Checklist:

- [ ] I have either updated the source of truth or arranged for update with product owner if needed https://breedinginsight.atlassian.net/wiki/spaces/BI/pages/1559953409/Source+of+Truth

[BI-2848]: https://breedinginsight.atlassian.net/browse/BI-2848?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/org/breedinginsight/api/model/v1/request/query/GenotypeImportQuery.java
+++ b/src/main/java/org/breedinginsight/api/model/v1/request/query/GenotypeImportQuery.java
@@ -1,3 +1,20 @@
+/*
+ * See the NOTICE file distributed with this work for additional information
+ * regarding copyright ownership.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package org.breedinginsight.api.model.v1.request.query;
 
 import io.micronaut.core.annotation.Introspected;
--- a/src/main/java/org/breedinginsight/api/model/v1/request/query/GenotypeImportQuery.java
+++ b/src/main/java/org/breedinginsight/api/model/v1/request/query/GenotypeImportQuery.java
@@ -1,0 +1,52 @@
+package org.breedinginsight.api.model.v1.request.query;
+
+import io.micronaut.core.annotation.Introspected;
+import lombok.Getter;
+import org.apache.commons.lang3.StringUtils;
+
+import java.util.ArrayList;
+import java.util.List;
+
+@Getter
+@Introspected
+public class GenotypeImportQuery extends QueryParams {
+
+    private String sampleSubmissionId;
+    private String projectNameForSampleSubmission;
+    private String sampleSubmissionCreatedBy;
+    private String genotypingFileName;
+    private String genotypingImportDate;
+    private String genotypingImportBy;
+
+    public SearchRequest constructSearchRequest() {
+        List<FilterRequest> filters = new ArrayList<>();
+
+        if (!StringUtils.isBlank(getSampleSubmissionId())) {
+            filters.add(constructFilterRequest("sampleSubmissionId", getSampleSubmissionId()));
+        }
+        if (!StringUtils.isBlank(getProjectNameForSampleSubmission())) {
+            filters.add(constructFilterRequest("projectNameForSampleSubmission", getProjectNameForSampleSubmission()));
+        }
+        if (!StringUtils.isBlank(getSampleSubmissionCreatedBy())) {
+            filters.add(constructFilterRequest("sampleSubmissionCreatedBy", getSampleSubmissionCreatedBy()));
+        }
+        if (!StringUtils.isBlank(getGenotypingFileName())) {
+            filters.add(constructFilterRequest("genotypingFileName", getGenotypingFileName()));
+        }
+        if (!StringUtils.isBlank(getGenotypingImportDate())) {
+            filters.add(constructFilterRequest("genotypingImportDate", getGenotypingImportDate()));
+        }
+        if (!StringUtils.isBlank(getGenotypingImportBy())) {
+            filters.add(constructFilterRequest("genotypingImportBy", getGenotypingImportBy()));
+        }
+
+        return new SearchRequest(filters);
+    }
+
+    private FilterRequest constructFilterRequest(String field, String value) {
+        return FilterRequest.builder()
+                .field(field)
+                .value(value)
+                .build();
+    }
+}
--- a/src/main/java/org/breedinginsight/api/v1/controller/geno/GenotypeDataUploadController.java
+++ b/src/main/java/org/breedinginsight/api/v1/controller/geno/GenotypeDataUploadController.java
@@ -6,18 +6,25 @@ import io.micronaut.http.annotation.*;
 import io.micronaut.http.multipart.CompletedFileUpload;
 import lombok.extern.slf4j.Slf4j;
 import org.brapi.client.v2.model.exceptions.ApiException;
-import org.breedinginsight.api.auth.AuthenticatedUser;
-import org.breedinginsight.api.auth.ProgramSecured;
-import org.breedinginsight.api.auth.ProgramSecuredRole;
-import org.breedinginsight.api.auth.SecurityService;
+import org.breedinginsight.api.auth.*;
+import org.breedinginsight.api.model.v1.request.query.QueryParams;
+import org.breedinginsight.api.model.v1.response.DataResponse;
 import org.breedinginsight.api.model.v1.response.Response;
+import org.breedinginsight.api.model.v1.validators.QueryValid;
 import org.breedinginsight.api.v1.controller.metadata.AddMetadata;
 import org.breedinginsight.brapps.importer.model.response.ImportResponse;
+import org.breedinginsight.model.GenotypeImportDetails;
+import org.breedinginsight.model.Program;
+import org.breedinginsight.services.ProgramService;
 import org.breedinginsight.services.exceptions.AuthorizationException;
 import org.breedinginsight.services.exceptions.DoesNotExistException;
 import org.breedinginsight.services.geno.GenotypeService;
+import org.breedinginsight.utilities.response.ResponseUtils;
+import org.breedinginsight.utilities.response.mappers.GenotypeImportQueryMapper;
 
 import javax.inject.Inject;
+import javax.validation.Valid;
+import java.util.Optional;
 import java.util.UUID;
 
 @Slf4j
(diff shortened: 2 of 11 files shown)


src/main/java/org/breedinginsight/api/model/v1/request/query/GenotypeImportQuery.java
```diff
@@ -0,0 +1,52 @@
+package org.breedinginsight.api.model.v1.request.query;
```
nickpalladino, line 18: Missing file header

src/main/java/org/breedinginsight/api/v1/controller/geno/GenotypeDataUploadController.java
nickpalladino, line 17: Can we also add a header to this file

src/main/java/org/breedinginsight/model/GenotypeImportDetails.java
```diff
@@ -0,0 +1,30 @@
+package org.breedinginsight.model;
```
nickpalladino, line 18: Missing file header

src/main/java/org/breedinginsight/utilities/response/mappers/GenotypeImportQueryMapper.java
```diff
@@ -0,0 +1,37 @@
+package org.breedinginsight.utilities.response.mappers;
```
nickpalladino, line 18: Missing file header

src/main/resources/db/migration/V1.36.0__create_genotype_import_table.sql
```diff
@@ -0,0 +1,55 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+-- Join table linking sample submissions to genotype imports
+create table genotype_import
+(
+    like base_entity including defaults including constraints including indexes,
+    sample_submission_id uuid not null,
```
nickpalladino, line 22: One side effect of linking the sample submission with the import is that if a sample submission with an associated genotype import is deleted we would now get a database error. I added a card to address this in a future sprint: https://breedinginsight.atlassian.net/browse/BI-2936

src/main/java/org/breedinginsight/services/geno/impl/GigwaGenotypeServiceImpl.java
```diff
@@ -266,6 +274,51 @@ public GermplasmGenotype retrieveGenotypeData(UUID programId, UUID germplasmId)
         }
     }
 
+    @Override
```
nickpalladino, line 273: I think a better separation of concerns here would be to move these methods out of the service into a GenotypeImportDAO similar to what we do in other areas.

src/main/java/org/breedinginsight/services/geno/GenotypeService.java
```diff
@@ -6,11 +6,15 @@
...
 
+import java.util.List;
 import java.util.UUID;
 
 public interface GenotypeService {
     ImportResponse submitGenotypeData(UUID userId, UUID programId, UUID submissionId, CompletedFileUpload uploadedFile) throws DoesNotExistException, AuthorizationException, ApiException;
 
     GermplasmGenotype retrieveGenotypeData(UUID programId, UUID germplasmId) throws DoesNotExistException, AuthorizationException, ApiException;
+
+    List<GenotypeImportDetails> getGenotypeImports(UUID programId);
```
nickpalladino, line 35: I think ideally we would want this in a separate service because it's not specific to a particular implementation of a GenotypeService. However since there is currently coupling in the gigwa service implementation it would require a larger refactor that I don't think we should take on now. Could be a future improvement, particularly if we ever want to support a different service implementation like the brapi prod server.

src/main/java/org/breedinginsight/services/geno/impl/GigwaGenotypeServiceImpl.java
```diff
@@ -266,6 +274,51 @@ public GermplasmGenotype retrieveGenotypeData(UUID programId, UUID germplasmId)
...
+                        .sampleSubmissionCreatedBy(record.get(sampleSubmissionCreatedByUser.NAME))
+                        .genotypingFileName(record.get(IMPORTER_IMPORT.UPLOAD_FILE_NAME))
+                        .genotypingImportDate(record.get(IMPORTER_IMPORT.CREATED_AT))
+                        .genotypingImportBy(record.get(genotypingImportByUser.NAME))
+                        .build());
+    }
+
+    private void createGenotypeImportLink(UUID submissionId, UUID importerImportId, UUID userId) {
+        OffsetDateTime now = OffsetDateTime.now();
+        log.debug("Inserting record into GenotypeImport table for submissionId={}, importerImportId={}, userId={}", submissionId, importerImportId, userId);
```
nickpalladino: Probably can remove the debug log statements here and method above

src/main/java/org/breedinginsight/services/geno/impl/GigwaGenotypeServiceImpl.java
```diff
@@ -266,6 +274,51 @@ public GermplasmGenotype retrieveGenotypeData(UUID programId, UUID germplasmId)
...
+        dsl.insertInto(GENOTYPE_IMPORT)
```
nickpalladino: You could also use the insert method generated on the entity jooq generates but don't have to change anything, just wanted to let you know that it's an option that we use in other places.

src/main/java/org/breedinginsight/services/geno/impl/GigwaGenotypeServiceImpl.java
```diff
@@ -266,6 +274,51 @@ public GermplasmGenotype retrieveGenotypeData(UUID programId, UUID germplasmId)
         }
     }
 
+    @Override
+    public List<GenotypeImportDetails> getGenotypeImports(UUID programId) {
```
nickpalladino, line 274: A common pattern we've used in other areas is to extend the jooq generated entity for the model and add a `parseSQLRecord(Record record)`. I don't think you need to change this but just wanted you to be aware of it.

src/test/java/org/breedinginsight/services/geno/impl/GigwaGenotypeServiceImplIntegrationTest.java
```diff
@@ -409,31 +405,87 @@ public void testSubmitValidFile() throws IOException, ApiException {
...
 
         ImportResponse response = importResponse.get();
         assertNotNull(response);
         assertNotNull(response.getProgress());
-        assertEquals((short)HttpStatus.ACCEPTED.getCode(), response.getProgress().getStatuscode(), "Error importing geno file: " + response.getProgress().getMessage());
+        assertEquals((short) HttpStatus.ACCEPTED.getCode(),
+                response.getProgress().getStatuscode(),
+                "Error importing geno file: " + response.getProgress().getMessage());
+
+        Integer joinRowCount = dsl.fetchOne(
```
nickpalladino: Instead of using the dsl and doing sql here you could use the `fetchBySampleSubmissionId` method in the GenotypeImportDAO. In general if there is a DAO method I think we should prefer that over raw sql in tests.

Also since the preexisting tests in here use mocks instead of database level checks I think it would be more clear to maintain that pattern rather than have a mix of mocks and db state checks.

nickpalladino (changes requested):
Nice job and thorough tests. I left a few minor coding standard comments and had a couple suggestions on design and tests.

src/main/java/org/breedinginsight/api/model/v1/request/query/GenotypeImportQuery.java
humsika: Added the header as per suggestion.

src/main/java/org/breedinginsight/api/v1/controller/geno/GenotypeDataUploadController.java
humsika: Added the header as per suggestion.

src/main/java/org/breedinginsight/model/GenotypeImportDetails.java
humsika: Added the header as per suggestion.

src/main/java/org/breedinginsight/utilities/response/mappers/GenotypeImportQueryMapper.java
humsika: Added the header as per suggestion.

src/main/resources/db/migration/V1.36.0__create_genotype_import_table.sql
humsika: I agree that we would see an error in that scenario, thanks @nickpalladino for creating a story card to address this.

src/main/java/org/breedinginsight/services/geno/impl/GigwaGenotypeServiceImpl.java
humsika: Thanks for the suggestion Nick, and I agree that it's better to separate the code and put it into the concerned classes. I updated the logic to reflect the same.

src/main/java/org/breedinginsight/services/geno/GenotypeService.java
humsika: I agree, but since we were using the existing controller I went with the current service classes. Buts its better to keep it separate as the API can grow independently.

src/main/java/org/breedinginsight/services/geno/impl/GigwaGenotypeServiceImpl.java
humsika: Thanks, removed as per suggestion.

src/main/java/org/breedinginsight/services/geno/impl/GigwaGenotypeServiceImpl.java
humsika: Thanks for pointing it out, I updated the code so it follows the standard practice in our team.

src/main/java/org/breedinginsight/services/geno/impl/GigwaGenotypeServiceImpl.java
humsika: Thanks for pointing it out, I updated the code so it follows the standard practice in our team.

src/test/java/org/breedinginsight/services/geno/impl/GigwaGenotypeServiceImplIntegrationTest.java
humsika: Thanks for the suggestion Nick, since the test was an integration test I went with the direct db lookup approach rather than just mocking like we do in unit/mock tests. I updated the new tests in this class to follow the existing mock based approach.

nickpalladino (approved):

